PR #2, opened 2018-05-02
Feat/projects states progress

dpnova:
Need to add graphql still.


web/web/settings/local.py
```diff
@@ -1,7 +1,7 @@
-from web.settings.base import *  # noqa
 from os import environ
-
 SECRET_KEY = environ['DJANGO_SECRET_KEY']
```
cballantyne: Was it intentional to change ordering of imports and put SECRET_KEY before from web.settings.base?

web/user_accounts/models.py
```diff
@@ -48,4 +48,9 @@ class User(AbstractUser):
     objects = CustomUserManager()
 
     def __str__(self):
-        return self.email
+        """
```
cballantyne: DB has email as not null so this should not happen. Is this left over from testing eth login?

PR #3, opened 2018-05-03
Ethereum Login

fmgoncalves:
Ethereum-based login

#### Issues/TODOs
 - The double panel UI on login isn't exactly pretty or friendly
 - When registering with Ethereum login, the user that gets created has no other info associated (email, password, first/last name, etc)
 - Ethereum address association for existing user is present at `/eth/setup` (for logged in users with no eth address exclusively) but isn't linked from anywhere - aybe show current address in profile and link from there?
 - Router path naming [is awful](https://martinfowler.com/bliki/TwoHardThings.html)
 - The ethereum session challenge is an in-memory dict, so it will be completely cleared at every server restart


dpnova:
This is looking great thanks mate!

PR #4, opened 2018-05-04
fix security group for ec2 instance.

cballantyne:


cballantyne:
Moved to individual aws_security_group_rule records as get cycle error when 2 groups trying to reference each other. 

Moved them all as terraform gets confused when you have some in the ingress setting inside the aws_security_group and another separate aws_security_group_rue.

dpnova:
If you're comfortable with this going live please accept @cballantyne - looks ok to me

PR #6, opened 2018-05-09
Feat/cb aml v1

cballantyne:
Basic AML working.
Modified processing of KYC and AML responses from providers so can be chained together more easily.
Country select in form.
Replaced full name with first name and last name.


web/kyc/management/commands/process_netverify_callbacks.py
```diff
@@ -0,0 +1,78 @@
...
+                send_fail_email(scan.user.email)
+
+                self.stdout.write(
+                    self.style.ERROR(
+                        f'Callback "{callback.id}" - KYC failed. Email sent.'
+                    )
+                )
+
+            callback.processed = True
+            callback.date_updated = datetime.datetime.now(pytz.timezone("UTC"))
```
cballantyne: @dpnova - is there a better way to do datetime UTC in python?

dpnova: `datetime.now(pytz.utc)`

dpnova:
This is all looking good man. Let's hold on this until we're ready to convert to the newer lambda stuff. We'll pair on converting some of it over then..

PR #14, opened 2018-05-11
Basic Modifications under Update Profile to test whitelist smart cont…

troowala:
# Update platform/web/user_accounts/templates/user_accounts/profile.html with id_ethaddress 

# Update update_profile function in platform/web/user_accounts/views.py with addAddressToWhitelist(ethaddress) functionality

# Copy whitelist.json consisting of the ABI/JSON for whitelist contract to platform/web/user_accounts folder

# Define WALLETPASS environment variable in docker-compose.yml of wallet owning the contract in the geth instance.


dpnova:
Great, good to prove this works. We'll have a think on the best way to integrate this properly now. Leaving open now so we can discuss... thanks mate!

troowala:
Thanks Felipe, Quite insightful feedback and i would like to understand better how we can implement it and also have some questions around the points you raised.

1. Currently the field/column eth_address is under table user_accounts_ethereum , however the rest of the profile fields are under the table  user_accounts_profile , the current update profile function will need significant modifications reading and writing to 2 different tables from my limited understanding it would be neater if the eth_address was mapped to a field under user_accounts_profile. Is this a mod we can make ?

2. Yep calling any write functions to a contract is definitely not a wise idea however the reality of an ethereum smartcontract platform as you know is this is something that the ES platform needs to be able to do securely. So the beneficiary wallet is kept separate from the owner/publisher wallet that interacts with the contract and the owner/publisher wallet has minimal funds in it to cover the gas costs. Also all our published contracts can be monitored using a backend python script (im tinkering with this currently) that ensures that variables such as beneficiary address or owners are changing on the contracts and if so then an email is immediately sent out.  Alternatively we can explicitly set the wallet that resides on the geth node to only be able to do addaddresstowhitelist() and nothing more. What are your thoughts on either of these methods ? 

3. The node allows RPC connections only from the django app server and the wallet is not always unlocked but only for 10 seconds for each write call being made using the walletpass environment variable. To access this vulnerability the django app or app server will need to be compromised in some manner and even then the attacker can only run away with some gas money or for a short period of time change the beneficiary wallet address to recieve any funds sent to the crowdsale from that point on and this will be detected immediately by the email notification. This is the best protection i could think of but im sure we can jam on this and collectively come up with a best method.

4. Agree , currently its just a .json file under the user_accounts folder but we could also look at publishing the ABI's to a gist since we are going opensource anyways.

PR #15, opened 2018-05-11
Fix crash on registration of existing user

fmgoncalves:
When a user attempted to register an email address of a user that already existed, the code crashed and the user got presented a 500 rather than a validation error in the registration screen.

On an implementation note, I'm not particularly happy about adding db access inside form validation, but it does make the code a bit safer to use (it becomes harder to introduce human error when using `RegistrationForm`) and makes this scenario testable.


dpnova:
Deployed and tested successfully.

PR #16, opened 2018-05-12
Feature/ssot

fmgoncalves:


dpnova:
I'm accepting now so we can start testing live. I've tested locally and it's working nicely..

PR #22, opened 2018-05-16
Use manifest files.

dpnova:


web/runserver_prod.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+pip install -r requirements.txt
```
cballantyne: I don't think this is required. The requirements are already installed in Dockerfile.

PR #25, opened 2018-05-16
Fixes and improvements to Announcements/SSOT

fmgoncalves:
Fix a signing bug when there is a line feed char in content. This bug is due to a mismatch between toHex functions on web3py and web3js.

Minor improvements around usability (no login required for announcements page, focus on content, etc)


web/ssot/templates/announcements.html
```diff
@@ -3,7 +3,17 @@
 
 {% block content %}
 
+{% comment %}
+Ugly TEMPORARY hack follows
```
dpnova: I can wear comments like this but I'd rather it describe when it needs to be revisited...

dpnova:
I'm thinking we should render out the addresses and ticks in a list? They are getting split oddly atm... what do you reckon @ejjick  ?

ejjick:
Just tested it. Works, but new lines become visible `<br />`, e.g "Ipsum?<br />Lorem" (we should allow HTML perhaps)

web/ssot/templates/announcements.html
```diff
@@ -3,7 +3,17 @@
...
+{% endcomment %}
+<style>
```
ejjick: Inline CSS is a no-no in my book. I can change our code to hide the menu when not logged in though.

ejjick:
> I'm thinking we should render out the addresses and ticks in a list? They are getting split oddly atm... what do you reckon @ejjick ?

I agree @dpnova Got a link to an announcement with 2+ signatures that I can preview to fix the design?

dpnova:
Feel free to pull and make the changes @ejjick

ejjick:
Just pushed some updates.

Let's discuss riot / code style / prototyping vs prod at some point... I agree with having the `ssot-msg.tag` template (it's generated many times and lots of JS going on). I don't agree with having `ssot-header.tag` (only generated once, nothing happening client-side from what I can tell).

dpnova:
I'm going to veto this and say we keep it. :P

dpnova:
ah the header tag? yeah whatever... keep the body one though!

dpnova:
we got bigger balls to grab atm anyway

dpnova:
Also hello world - welcome to our github :)

dpnova:
No breaking db changes here so I'm accepting.

PR #27, opened 2018-05-17
Feat/welcome page

ejjick:
Welcome page after the user has logged in.

(also fixed the main navigation so that it's hidden until the user is logged in)


dpnova:
Hey man can you add a unit test for that basic index method? Should be trivial one...

dpnova:
@ejjick did you need a hand here or want me to do it?

PR #28, opened 2018-05-17
Feat/cb kyc new status page

cballantyne:
Cleaned up status page.
Cleaned up urls for various pieces, ie netverify and docverify.
Glue together different processes that make up KYC process.
Profile page displays verified status.
ID Check verifies name and dob.
Doc check verifies name and address.


web/kyc/docverify.py
```diff
@@ -39,3 +39,18 @@ def get_unprocessed_callbacks(self):
...
+    def get_latest_callback(self, user_id):
+        return DocverifyCallback.objects.filter(
+            user_id=user_id
+        ).order_by('created').first()
+
+    def get_verification_status(self, user_id):
+        docverify_processed_status = 'NOT_DONE'
+        latest_callback = self.get_latest_callback(user_id)
+        if latest_callback is not None:
+            docverify_processed_status = 'PROCESSING'
```
dpnova: I notice you're using const's later.. probably makes sense to have them here?

cballantyne: Yeah you are right. I will put a TODO to clean these up. Already have a tweak or two to do after deploying.

PR #32, opened 2018-05-23
Token added to netverify callback.

cballantyne:
Adding security to jumio callback urls.


web/kyc/netverify.py
```diff
@@ -51,10 +51,10 @@ def get_authorization_token(self, request_params, app_version):
                 "Netverify fetch authorization token failed"
             )
 
-    def store_scan_data(self, request):
+    def store_scan_data(self, request, id):
         params = request.GET
         scan = NetverifyScan(
-            id=uuid.uuid4(),
+            id=uuid.UUID(id),
```
cballantyne: Here I was able to use the id I generated previously as the one saved in the db. However I could not do this later on for Docverify because the id was not passed back by Jumio.

web/kyc/views.py
```diff
@@ -105,11 +125,22 @@ def status(request):
...
+    # Compare token with hash of scan_id to make sure valid callback
+    token_to_compare = create_token(scan_id)
+
+    if token_to_compare != token:
+        logger.debug(f'token did not match {token_to_compare}')
+        return HttpResponseForbidden()
+    else:
+        netverify_client = get_netverify_client()
+        netverify_client.store_callback_data(request)
+        return JsonResponse({'status': 'error'})
```
ejjick: should be status "ok"

PR #33, opened 2018-05-25
Working version of management command to add address to whitelist con…

cballantyne:
Management command to add eth address to whitelist contract.
All pieces of info need to be entered but some of these will become optional as we add more pieces later like env var for geth_node.
This is just the first piece - add_address_to_whitelist that will be hooked up to other processes later.


troowala:
just watched the demo , the call and writing was sucessful , looks good.

PR #40, opened 2018-05-30
Kyc tweaks - welcome and profile

cballantyne:


web/kyc/templates/kyc/docverify_select_type.html
```diff
@@ -3,7 +3,16 @@
...
             <div class="col col-md-8">
-                <h1 class="h1-responsive text-center">Get verified - Upload Document</h1>
+                <h1 class="h1-responsive text-center">Proof of address</h1>
+
+                <div class="alert alert-warning alert-dismissible fade show" role="alert">
+
+                    <p>In the next step of the verification process, you'll need to submit a proof of address document like a bank statement or utility bill. This must be <strong>no more than 3 months old</strong> and clearly state your <strong>full name</strong>, <strong>address</strong> and <strong>issue date</strong>.</p>
+
+                    <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+                        <span aria-hidden="true">&times;</span>
```
ejjick: There's a Font Awesome icon for this... best if we use them. Trivial for now though...

https://fontawesome.com/icons/times?style=solid

`<i class="fas fa-times"></i>`

web/user_accounts/models.py
```diff
@@ -100,6 +100,7 @@ class Profile(TimeStampedModel):
     verified = models.BooleanField(default=False)
     suburb = models.CharField(max_length=200, null=True, blank=True)
     eth_address = models.CharField(max_length=42, null=True, blank=True)
+    email = models.EmailField(max_length=300, null=True, blank=True)
```
ejjick: If the email is nullable we'll have trouble emailing the user about KYC :) Sounds like you're onto it though...

web/kyc/templates/kyc/docverify_select_type.html
cballantyne: Good to know. Thanks. 
For this I don't think relevant since it is using Bootstrap Alert (https://getbootstrap.com/docs/4.0/components/alerts/#dismissing). But I could probably use elsewhere icons like in Profile screen.

web/user_accounts/models.py
cballantyne: Yeah I realised yesterday this would be an issue and need to get onto it after this is merged in. Someone could log in with their eth address and never supply and email ... in which case we don't send them any.

PR #41, opened 2018-05-31
Connect fuzzy as default matcher

cballantyne:
Default matcher return is now FuzzyMatcher which allows for a bit of wriggle room in specifying addresses. I have added a bunch of extra tests around this as well.
FuzzyMatcher should take care of cases where address is extracted without trailing comma.
We are still matching exact on first_name (extracted name starts with), last_name (extracted name ends with) and postcode.


web/kyc/management/commands/process_docverify_callbacks.py
```diff
@@ -115,9 +115,11 @@ def handle(self, *args, **options):
...
-
         if profile is None:
             return False
 
-        matcher = get_default_matcher(profile, extracted_data)
-        return matcher.matches()
+        try:
+            matcher = get_default_matcher(profile, extracted_data)
+            return matcher.matches()
+        except RuntimeError:
```
dpnova: Can you add a log message here, this could be surprising and frustrating error to track down and debug.

web/kyc/matching.py
```diff
@@ -100,34 +122,59 @@ def matches(self) -> bool:
...
+        return self.first_name() \
+            and self.last_name() \
+            and self.postal_code() \
+            and self.match_score() > 90
+
+    def match_score(self) -> bool:
+        p_addr = self.create_profile_addr()
+        e_addr = self.create_extracted_addr()
+        distance_score = fuzz.partial_token_sort_ratio(p_addr, e_addr)
+        print(f'fuzzy comparing addresses {p_addr} with {e_addr}')
```
dpnova: Just need to ditch the prints..

web/kyc/management/commands/process_docverify_callbacks.py
cballantyne: Yeah agreed. Done.

web/kyc/matching.py
cballantyne: Got rid of those. There were a lot more originally... must have missed those 2. haha.

PR #42, opened 2018-06-01
KYC Tweaks 2

cballantyne:


web/web/templates/web/index_not_verified.html
```diff
@@ -0,0 +1,35 @@
...
+            <p>
+                We’re excited to have you join the community.
+            </p>
+            <p>
+                You can use the platform to discover, invest and contribute to interesting ICO projects or create your own.  
+            </p>
+            <p>
+                Take some time to explore the different features.
+            </p>
+            <p>
```
ejjick: IMHO, at some point, these three paragraphs should just be an include file triggered by an if-statement.

web/web/views.py
```diff
@@ -1,8 +1,19 @@
...
 def index(request):
-    return render(request, 'web/index.html')
+    # check if user is verified
+    profile_verified = False
+    if Profile.objects.filter(user_id=request.user.id).exists():
+        profile = Profile.objects.get(user_id=request.user.id)
+        if profile.verified:
+            profile_verified = True
+
+    if profile_verified:
```
ejjick: Replace with `return render(request, 'web/index.html', {profile_verified: profile_verified})` and slight modification to `index.html`?

cballantyne: Agree with this. But I wasn't sure what the content would be at this stage. I only guessed the content might be quite different and have dashboard type stuff later. That was my only reasoning for doing it this way with separate html file.

web/web/templates/web/index_not_verified.html
cballantyne: Yeah totally. All depends on the content in the two index pages and how much they differ.

PR #49, opened 2018-06-13
Feat/project details

ejjick:
TSE landing page, generated from DB with templates for markdown and team/people sections. Discount countdown timer for the different stages.


web/tse/tests/test_views.py
```diff
@@ -94,3 +98,43 @@ def test_authz_denied_public(self):
...
+        discount_stage_b = DiscountStage(project=self.project, created_by=creator, title='Stage B', discount=25, start_date=datetime.now(pytz.UTC)-less_two_days, end_date=datetime.now(pytz.UTC)-less_three_days)  # noqa: E501
+        discount_stage_b.save()
+
+        page_section_markdown = ProjectPageSection(project=self.project, created_by=creator, template='markdown', data="I'm [Markdown](https://www.markdown.org/) enabled with **bold** and _italic_ styling.", order_nr=10)  # noqa: E501
+        page_section_markdown.save()
+
+        page_section_people = ProjectPageSection(project=self.project, created_by=creator, template='people', data="- name: John Doe\n  image: https://path.to/img/john.png\n  title: CEO\n  linkedin: https://www.linkedin.com/in/johndoe", order_nr=20)  # noqa: E501
+        page_section_people.save()
+
+    def test_view_public(self):
```
cballantyne: This should be called 'test_project_details_view'?

web/tse/tests/test_views.py
```diff
@@ -94,3 +98,43 @@ def test_authz_denied_public(self):
...
+        request = self.factory.get('/tse/project/' + str(self.project.id))
+        request.user = auth.get_user(Client())
+
+        self.assertTrue(request.user.is_anonymous)
+
+        response = project_details_view(request, self.project.id)
```
cballantyne: I think we could test a bit more here. For example in that view you are doing some looping and building objects that are stored in the context. So test those objects are as expected, eg size, content.

```
context={
            'project': project,
            'discount_stages': discount_stages,
            'current_discount': current_discount,
            'sections': sections
        }
```

web/tse/templates/tse/project_details.html
```diff
@@ -0,0 +1,120 @@
...
+	$("#subnav li a[href^='#']").on('click', function(e) {
+		e.preventDefault();
+		var hash = this.hash;
+		$('html,body').animate({scrollTop: $(hash).offset().top}, 300, function(){window.location.hash = hash;});
+	});
+
+	{% if current_discount %}
+		// countdown timer
+		var countdownDate = new Date("{{ current_discount.end_date | date:'M j, Y H:i:s' }} UTC").getTime();
+		var x = setInterval(function() {
```
cballantyne: Not a blocker. Just an observation. We should figure out how to unit test the javascript that is growing in our platform. I have some floating around as well that could use some tests. I have used Mocha in the past. I think Riot supports Mocha tests or perhaps it has a better way.
 @dpnova @fmgoncalves ^^

web/tse/models.py
```diff
@@ -197,6 +198,46 @@ def __str__(self):
...
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    created_by = models.ForeignKey(User, on_delete=models.DO_NOTHING)
+    project = models.ForeignKey(
+        Project,
+        on_delete=models.DO_NOTHING,
+        null=True
+    )
+    title = models.CharField(max_length=200)
+    discount = models.FloatField(null=True, blank=True)
```
fmgoncalves: We could probably use a little [validator](https://docs.djangoproject.com/en/2.0/ref/validators/) here. At least to ensure the discount is always between 0 and 100.

web/tse/templates/tse/project_details.html
fmgoncalves: I see the javascript in the platform as falling into three categories, each with its own testing strategies (IMO):
 1) Third party library: we need to trust this code to some degree, and therefore if we want to add tests to these, we should do it upstream.
 2) Medium to large blocks of application logic: these should be factored out into their own contained libraries (or replaced by equivalent third party lib) and be tested in isolation with jasmine/mocha/ava/etc. I've personally used jasmine and mocha and don't have any preference or strong feelings toward any.
 3) Small snippet/usage of (thid party or own) library: these are usually tighly coupled with the web page itself so I think that testing them indirectly though selenium/webdriver tests of the page's behavior should be enough provided the libs used are well tested on their own.

The lines below, for example fall somewhere between 2) and 3), as it seems like a good candidate to use something like [moment](https://momentjs.com/) or extract the time diff module out.

web/tse/views.py
```diff
@@ -102,3 +110,44 @@ def projects_view(request):
...
+        'order_nr'
+    ).filter(project=project)
+    allowed_tags = bleach.sanitizer.ALLOWED_TAGS + []
+    cleaner = Cleaner(tags=allowed_tags)
+    sections = []
+    for db_section in db_sections:
+        section = {}
+        section['title'] = db_section.title
+        if db_section.template == 'people':
+            people_data = yaml.load(db_section.data)
```
fmgoncalves: Given that `db_section.data` is a TextField on the database, and can contain malformed content, `yaml.load` here can crash, or worse, it can give off gibberish data that will cause a gibberish page to be rendered.

A bit of exception handling and some sort of sanity check (like verifying people_data, which gets fed into the renderer, is an array, and not a string or a number) is needed here.

web/tse/templates/tse/project_details.html
cballantyne: Yeah agreed about the little snippets -  they can be covered by selenium. The type 2 ones are the problem we need to figure out. I will put a ticket in to get to figure out how we can do it.

web/tse/tests/test_views.py
ejjick: requirements changed last minute, used to be both private and public views

web/tse/migrations/0006_merge_20180613_0645.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.0.5 on 2018-06-13 06:45
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tse', '0005_auto_20180613_0130'),
+        ('tse', '0004_projectwhitelist'),
```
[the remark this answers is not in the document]
ejjick: merged, thanks

ejjick:
Doing the validation and tests now... I'm sure we'd forget if I didn't, but the purpose of this project is just for us to be able to create these TSE pages internally. It is not being released to any users at this point.

web/tse/templates/tse/project_details.html
ejjick: Totally agree, just not something I feel is needed for this PR. At this point the code is only being used internally by Johan to create the TSE pages.

web/tse/views.py
ejjick: Thanks, have added some exception checking and validation. The tool is just for internal use so far so don't want to go over board with it.

I find it literally impossible to bread the markdown parsing so ignoring that.

ejjick:
All done, please have another look @cballantyne @fmgoncalves

fmgoncalves:
lgtm :+1:

web/tse/templates/tse/project_details.html
cballantyne: Yes, it is not needed for this PR. I have created a separate ticket. Not a blocker.

PR #50, opened 2018-06-14
Feat/localstack

ejjick:
Localstack docker integration for local dev with SNS, SQS, S3, and Lambda


web/docker-compose.yml
```diff
@@ -53,3 +53,17 @@ services:
...
+      - "4567-4583:4567-4583"
+      - "${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}"
+    environment:
+      - SERVICES=s3,sns,sqs,lambda
+      - DEBUG=1
+      - LAMBDA_EXECUTOR=docker
+      - PORT_WEB_UI=8080
+      - DOCKER_HOST=unix:///var/run/docker.sock
+    volumes:
+      - "${TMPDIR:-/tmp/localstack}:/tmp/localstack"
```
cballantyne: It bombed out for me running locally due to mount error. Works when I remove TMPDIR volume mapping. Is it required?

ejjick: May have something to do with this:

https://github.com/localstack/localstack#running-in-docker

> (Note that on MacOS you may have to run TMPDIR=/private$TMPDIR docker-compose up if $TMPDIR contains a symbolic link that cannot be mounted by Docker.)

Try setting the do you mind trying `export TMPDIR=/private/tmp/localstack`?

ejjick:
I just commented out `- "${TMPDIR:-/tmp/localstack}:/tmp/localstack"` and ran all of my validation tests. Everything works so I'm committing it with the volume config commented out (so that we get a reminder and can easily restore if something is acting weird).

PR #51, opened 2018-06-20
Sign public profile

cballantyne:
Ability for users to sign public profile of another user.


web/wot/models.py
```diff
@@ -13,6 +13,35 @@ class Profile(TimeStampedModel):
...
+{separator}'''.format(
+            eth_address=self.clean_for_sig(self.eth_address),
+            first_name=self.clean_for_sig(self.user.profile.first_name),
+            last_name=self.clean_for_sig(self.user.profile.last_name),
+            separator='='*20
+        )
+
+    # Metamask and web3py signatures do not match if
+    # LF or some other chars appear on the text, and this
+    # behaviour is breaking signing on some announcements.
```
dpnova: s/announcements/profiles

cballantyne: Oops. Fixed.

PR #53, opened 2018-06-21
KYC required decorator

ejjick:
KYC is required for organisations, projects, questionnaires, trading, and ssot. This decorator makes it easy to enforce it on views.


web/kyc/decorators.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def kyc_required(function):
+    def wrap(request, *args, **kwargs):
+        if (
+            request.user
+            and
+            Profile.objects.filter(user_id=request.user.id).exists()
+        ):
+            profile = Profile.objects.get(user_id=request.user.id)
```
dpnova: I would do this differently @ejjick @cballantyne - do the filter outside and force the queryset to resolve to a list. Then check it from there - doing it this way requires two queries in the success case.

PR #61, opened 2018-06-26
Feat/project invest

ejjick:
Due tomorrow 9am so trying to push this through. Got all details in so that it can be previewed. Just gotta add real time ETH-to-USD lookup with caching.


dpnova:
@ejjick sorry to be an ass - can you just cache the api call to coin market cap? Not the whole page? Bit wary of caching full pages when people are logged in.

ejjick:
Done. I used Django's "Template Fragment Caching" and set it to 5 minutes (that's how often Coin Market Cap updates).

dpnova:
approved

PR #65, opened 2018-07-03
project create flow

cballantyne:


web/tse/templates/tse/project_form.html
```diff
@@ -0,0 +1,79 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+
+<div class="container pt-4">
+    <div class="row justify-content-center">
+        <div class="col-md-6">
```
ejjick: Trivial, but I'd use `col col-lg-8 col-xl-6` here

PR #66, opened 2018-07-04
Feat/ssot v1

ejjick:
Features to allow users to update and re-sign announcements. Multiple presentation updates.


web/ssot/static/components/ssot-header.tag
```diff
@@ -1,41 +0,0 @@
-<token-header>
```
cballantyne: Can see you have replaced this with python include. Makes sense to me. Keep it simple.

PR #67, opened 2018-07-05
terraform configs

ejjick:
terraform configs for redirect of blockchainclub.io to enhancedsociety.com


infrastructure/corporate/main.tf
```diff
@@ -1,6 +1,6 @@
 variable "aws_region" {
   description = "The AWS region to create things in."
-  default     = "ap-southeast-2"
+  default     = "us-east-1"
```
cballantyne: This seems wrong. Should it not stay as ap-southeast-2?

infrastructure/corporate/main.tf
```diff
@@ -168,4 +156,74 @@ resource "aws_route53_record" "verify_blockchainclub_io" {
...
+  }
+}
+
+resource "aws_acm_certificate" "cert_blockchainclub_io" {
+  domain_name = "blockchainclub.io"
+  subject_alternative_names = ["*.blockchainclub.io"]
+  validation_method = "DNS"
+}
+
+resource "aws_cloudfront_distribution" "cloudfront_blockchainclub_io" {
```
cballantyne: I am confused why we need cloudfront setup. I thought it was just redirecting the domain. Also if we do need cloudfront I thought this would live under website/_infrastructure.

infrastructure/corporate/main.tf
ejjick: This is for the certificate, which is needed to redirect https links. Only N. Virginia is supported. Unfortunately you can't set a region for certs in TF.

https://www.terraform.io/docs/providers/aws/d/acm_certificate.html
https://www.terraform.io/docs/providers/aws/r/acm_certificate.html

I also had some errors with the S3 bucket for the redirect, which needed to be in N. Virginia too. Can't recall exactly but I think it had to do with the apex domain and the route53 `alias` record.

infrastructure/corporate/main.tf
ejjick: You can't assign a certificate to S3 hosted sites, which is why CF is needed. Without it all bookmarked links with https wouldn't resolve.

ejjick:
Not sure if it needs a migration to be honest. All I've added is a S3 bucket for the redirect, a CF distribution to support HTTPS, and a cert to go with it (WP Engine supplied it before). The rest of the records were there since before. The default region would be the only concern, but only in case we add new services as what's there right now either requires N. Virginia (S3, Cert) as the region or are Global (CF, Route53).

dpnova:
I guess I'm worried that we'll leave orphaned things in one region with a straight switch over like this?

If it's not a problem then I'l happy...

ejjick:
Yeah, I agree. Could be easy to forget. I'll give it some more thought. Would be good to get a better structure of the TF files. Perhaps we should just move these things to a separate config for just BCC.

infrastructure/corporate/main.tf
dpnova: yeah this is for the main website... not app.

dpnova:
Surely we can just do this now and remove any crap that's left over if TF doesn't get it right?

ejjick:
Just letting you know I'm moving this to the website `_infrastructure` and canning this PR, once SI is done.

PR #70, opened 2018-07-10
Project declaration requirement

cballantyne:
Moved country restriction to decorator.
Added project declaration requirement before viewing invest page for a project. This is enforced using decorator.


web/tse/decorators.py
```diff
@@ -0,0 +1,71 @@
...
+
+def tse_project_declaration_required(function):
+    def wrap(request, *args, **kwargs):
+        project_id = kwargs['pk']
+        if request.user:
+            project_declarations = ProjectDeclaration.objects.filter(
+                user_id=request.user.id
+            )
+
+            if len(project_declarations) > 0:
```
dpnova: better to use exists() here.

web/tse/views.py
```diff
@@ -291,37 +349,30 @@ def send_new_project_email(project):
...
-
-    if profile.country is None or profile.country_from_id is None:
-        return False
-    if profile.country in excluded_countries or profile.country_from_id in excluded_countries:  # noqa: E501
-        return False
-
-    # TODO - AUS non sophisticated investor restriction
-
-    return True
+    if (
```
dpnova: Here it's better to do the filter outside the if/else and check if the list is empty or not to avoid doing two queries.

dpnova: Or maybe do a try/except on the get()

web/tse/decorators.py
```diff
@@ -0,0 +1,71 @@
...
+from django.http import HttpResponseRedirect
+from django.urls import reverse
+
+
+def tse_country_restricted(function):
+    def wrap(request, *args, **kwargs):
+        if (
+            request.user
+            and
+            Profile.objects.filter(user_id=request.user.id).exists()
```
dpnova: Pontentially two queries here... do the profile filter outside and check if it's empty or not (or use a try/except on the get)

web/tse/views.py
cballantyne: Did try/except on get so only have 1 query now.

web/tse/decorators.py
cballantyne: Just one query now using filter and first(). Checking if profile is not None.

PR #71, opened 2018-07-18
Sophisticated investor

ejjick:
Sophisticated investor check. File upload interface for sophisticated investor documents. Uploads the file to an encrypted S3 bucket and provides approve/decline functionality in admin as well as direct links to the documents.

Caveats:
1. Deployment should be good but might be worth running in staging first
2. Local dev requires bucket creation (added to `runserver.sh`)
3. Code coverage is at 94.88% even though less is reported (codecov is down atm)


web/user_accounts/admin.py
```diff
@@ -54,3 +59,71 @@ class EthereumAccountAdmin(admin.ModelAdmin):
...
+            action_flag=CHANGE,
+            change_message=_('Declined')
+        )
+    self.message_user(
+        request,
+        "%s sophisticated investor document(s) declined" % affected_rows)
+    email_content = render_to_string(
+        'user_accounts/email/si_declined.txt'
+    )
+    send_mail(
```
dpnova: Can we not use fail silently here? I'd rather report something to sentry than have it gobbled up.

web/user_accounts/models.py
```diff
@@ -130,3 +134,25 @@ def completed(self):
             return {'completed': False, 'reason': 'eth_address is None'}
 
         return {'completed': True, 'reason': None}
+
+
+def rename_file(instance, filename):
```
dpnova: This will barf if the file has no extension.

web/user_accounts/admin.py
cballantyne: Fixed up to log an error ... which in turn gets reported to sentry.

web/user_accounts/models.py
cballantyne: It actually doesn't. I was surprised too. If no extension the new filename has no extension either. I have put a couple of tests around it to make sure working as expected.

ejjick: No extension makes the form invalid too. Thanks for testing though.

PR #75, opened 2018-07-20
Sign Ethereum transaction locally instead of  unlocking account on ethereum node

fmgoncalves:
Remove dependency on ethereum node to sign transactions, effectively removing running a private ethereum node as an operational dependency.

We still need to connect to a node to interact with the blockchain, but it can be any node, without caring about operational or security considerations.


fmgoncalves:
### DONE

- Replaced geth with ganache-cli as a testing node in docker compose.
- Made tests rely on ganache-cli configuration.
- Removed all PoA node references. Added benefit that node/chain config is independent of code.
- Replaced all contract owner address/password with contract owner private key.
- Added test utility methods to setup whitelist on test node if it doesn't exist already.
- Replaced calls to smart contract with account unlocked on eth node with on-app signing before broadcasting to the eth node.

### TODO

- Setup defaults for eth rpc stuff environment variables. Maybe connect to ropsten where a whitelist contract has been previously deployed or something.
- Validate implementation choices.

troowala:
since goal is to use signed tx's with public nodes , it seems bundling a local instance of ganache or geth is redundant. 

can we drop this and simply specify infura.io or ropstein.infura.io as default geth_node_url. this is because rinkeby & geth in --dev mode are both POA and removing the POA part of the code breaks compatibility with these networks.

troowala:
/code/tse/views.py needs the references to GETH_NODE_POA removed to sort out the error that occurs when clickin on invest

however the whitelist was still not written to.

troowala:
notice that buildtransaction when connected to ganache or geth in dev mode does not require gas to be specified explicitly however when connected to any of the testnets or mainnet the gas value needs specifying such as below.

txn.buildTransaction({
            'gas': 51000,
            'nonce': nonce,
        })

fmgoncalves:
Thanks for the review @troowala !!

I fixed the explicit gas setting as well as the leftover references (these were caused by merging master, and I missed it even though they broke tests :S).
On the other counts:

I replaced `geth` with `ganache` to make it **explicit** that the node included is for tests only. In the `geth_node_client_tests` (I didn't change the name) the `settings.GETH_NODE_URL` isn't used, `http://ganache:8545` is hardcoded instead. The tests need to be reproducible and work on as little legacy data as possible, hence them running on a blank test blockchain. For this to work, I also added a hardcoded private key, corresponding to the funded address on `ganache`, and a function that sets up the whitelist contract if it doesn't exist already.

I purposefully left the `GETH_NODE_URL` and `GETH_WHITELIST_OWNER_PRIVATE_KEY` blank, and expect the defaults to be set to ropsten.infura.io (because the default should be local dev and for that it makes sense to use a testnet) and some disposable private key funded through a faucet. `ganache` is just for tests, setting these environment variables should be what defines a development setup. This is also why `ganache`'s port isn't exposed.

Finally, PoA code is very specific to testing PoA networks and personally I think, given that PoW testnets exist and the mainnet is PoW as well, having PoA code and flags is unnecessary complexity.

troowala:
I tried this but it seems the whole write to whitelist part is being completely skipped. i can verify this because even if i dont put in any project whitelist it doesnt error out. @cballantyne is this because of the recent change  that was made to allow for the demo last friday ? can you please confirm what setting is it that needs to be changed so the write to whitelist part is not skipped. thx

troowala:
okay as per discussion have made the change so that the write to whitelist part is not skipped and it is rightly not being skipped however the action itself is not happening.

web_1         | [01/Aug/2018 02:23:06] "POST /tse/project/cadf8fc3-b397-445a-a679-2c96078c980c/declaration HTTP/1.1" 302 0
web_1         | DEBUG 2018-08-01 02:23:06,435 views 114 140115254966016 0x6641c30B8Ec57A6168707115b80B0af1a5cf160C NOT_ADDED to whitelist 0x92ce566566e7012140aa16e5ab30314d00a2b51c

fmgoncalves:
It fails only if the private key isn't prefixed with `0x`, but I will relax that requirement.

A successful example run is:

Django command for adding to whitelist
Checking on a block explorer the address was added


I only tested it through the django command, but the requirements are exactly the same for every usage:
  - make sure `GETH_NODE_URL` environment variable is set to an ethereum rpc node (geth, parity, whatever). On my example it was 'https://ropsten.infura.io'
  - make sure `GETH_WHITELIST_OWNER_PRIVATE_KEY` is set. On my example, it was set to the private key for the address `0x7c3d8b7ad75d77eb8734c7d4ae77074d92df3255`, which _had to be_ the [owner of the whitelist contract you want to interact with](https://ropsten.etherscan.io/address/0x78392814a5c6af62436170a9ab38de96d803f225#readContract).
  - make sure the provided whitelist contract address corresponds to a Whitelist contract _and_ that contract was created by the account corresponding to `GETH_WHITELIST_OWNER_PRIVATE_KEY`.

troowala:
Cool ! this works ...

also i assume for GETH_NODE_URL when you said https://ropsten.infura.io you meant https://ropsten.infura.io/v3/yourapikey here ? 

So now we just need now to sign up with an account with infura for ES with the same email used for signing up for our other API keys and declare this environment variable and we good to go.

fmgoncalves:
Nope, I meant  https://ropsten.infura.io, although that one shouldn't be used for any serious use and we should definitely create an account and get an API key.

troowala:
cool ... so @dpnova  are you going to sign up for the infura API key with your email and configure it for the GETH_NODE_URL value ? the GETH_WHITELIST_OWNER_PRIVATE_KEY i believe you already have access to in last pass ? apart from these environment variable prep on the system this PR looks good to be merged. Thanks

PR #78, opened 2018-07-30
Feat/dashboard updates

ejjick:
Dependent on the users state, show different content on the Dashboard. There are three different states.

Also implemented the "edit project" view.


web/tse/views.py
```diff
@@ -352,6 +352,26 @@ def project_create_view(request):
...
+@kyc_required
+def project_edit_view(request, pk):
+    try:
+        project = Project.objects.get(pk=pk, creator=request.user)
+    except Exception:
+        raise RuntimeError('Project could not be found')
+    if project:
+        if request.method == 'POST':
+            form = ProjectForm(request.POST, instance=project)
+            if form.is_valid():
```
cballantyne: I think we should have some validation that the project is owned by the user. 
Just after I typed that I noticed the get on line 359 uses creator as part of the filter as well. So that would take care of it? Could you please put a test to cover this explicit scenario - because it is very important to ensure is covered?

ejjick:
Additional test added @cballantyne I _was_ filtering by `creator` but never hurts to test twice in case things change.

PR #81, opened 2018-08-13
Event queues with localstack implementation

ejjick:


web/kyc/tasks.py
```diff
@@ -0,0 +1,146 @@
...
+    netverify_client = get_netverify_client()
+    callback = NetverifyCallback.objects.get(id=callback_id)
+    # lookup scan so can get user who we need to send email to
+    scan = netverify_client.get_scan_with_jumio_id_scan_reference(
+        callback.jumio_id_scan_reference
+    )
+    send_email = False
+    if scan is None:
+        callback.processed_status = NetverifyCallback.PROCESSED_STATUS_FAILED # noqa
+        # TODO - clean up this duplication.
```
cballantyne: Not great. But I don't want to complicate this PR too much so have put a TODO for this.

cballantyne:
I think this is ready to review. 
I have left out anything to do with running this in staging/production - this will come in separate PR once this one is merged in.

web/kyc/tasks.py
ejjick: Might be able to assign `output = logger` to control the output

cballantyne: I have fixed this up so tests don't depend on stdout. Those stdout writes now just go to logger. And the tests validate on what happened inside the method. I mock the send email methods to determine if they were called or not.

PR #90, opened 2018-10-04
expose organisation, shareholders, officeholders

ejjick:
GraphQL to query organisation, shareholders, officeholders.

Also realised we missed making `User` an optional relationship (for "claiming") in officeholder/shareholder models. Just set it to `blank=True`, so migration needed.


ejjick:
C's PR will bring up the coverage as he's got tests for `mutate` in `schema.py`

PR #93, opened 2018-10-06
graphql support for Add new shareholder and Create new sharetype with onchain option

troowala:
### Pull Request check-list

SE

### Description of change
Support for Add new shareholder
Support for Add new sharetype (for example ordinary , preference, etc)
Onchain Option (If set to true will create a token on chain with provided parameters)


web/se/contracts.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def create(self, token_quantity, token_name, token_symbol):
+        # Construct the TXN
+        constructed_txn = self.contract_.constructor(
+            token_quantity,
+            token_name,
+            token_symbol)
+
+        # Sign the TXN
+        tx_hash = TokenManager().sign(constructed_txn)
```
cballantyne: TokenManager() - this should be self?

web/se/contracts.py
```diff
@@ -0,0 +1,82 @@
...
+            abi=self.contract_interface['abi']
+        )
+
+        # Construct the TXN
+        constructed_txn = contract_instance.functions.transfer(
+            eth_address,
+            token_quantity*10**18)
+
+        # Sign the TXN
+        tx_hash = TokenManager().sign(constructed_txn)
```
cballantyne: TokenManager() - this should be self?

web/web/schema.py
```diff
@@ -94,12 +92,104 @@ class OrganisationInput(graphene.InputObjectType):
...
     organisation = graphene.InputField(OrganisationInput)
     shareholders = graphene.List(ShareholderInput)
     officeholders = graphene.List(OfficeholderInput)
 
 
+class ShareTypeInput(graphene.InputObjectType):
+    name = graphene.String(required=True)
+    symbol = graphene.String(required=True)
+    quantity = graphene.Int(required=True)
+    onchain = graphene.Boolean(required=False)
```
cballantyne: Not sure about exposing this here. We will discuss all the onchain stuff after these PRs. So happy to leave this for now.

PR #95, opened 2018-10-10
download & install solc

troowala:
### Pull Request check-list

web

### Description of change
Download & install solc


web/runserver.sh
```diff
@@ -7,6 +7,7 @@ fi
 python3 manage.py migrate
 sh run_cron_setup.sh
 sh run_celery_setup.sh
+sh run_solc_setup.sh
```
cballantyne: This probably needs to be in runserver_prod.sh as well? We have separate ones because of the s3 localstack diff for dev.

PR #97, opened 2018-10-10
Trigger asic update

ejjick:


ejjick:
I've fixed the conflict but wouldn't mind cleaning up the migration files if someone's keen to give us a hand

web/web/schema.py
```diff
@@ -244,8 +255,48 @@ def mutate(root, info, share_registry_data=None):
...
+
+
+class AsicUpdate(graphene.ObjectType):
+    organisation = graphene.Field(Organisation)
+    user = graphene.Field(User)
+
+
+class TriggerAsicUpdate(graphene.Mutation):
+    class Arguments:
+        organisation_id = OrganisationIdInput(required=True)
```
cballantyne: Could this be similar to other mutations? Just to keep consistent with other inputs. Correct me if I am being dumb.

asci_update_data = AsciUpdateInput(required=true)


class AsciUpdateInput(graphene.InputObjectType):
    organisation_id = graphene.ID(required=True)

web/web/schema.py
```diff
@@ -244,8 +255,48 @@ def mutate(root, info, share_registry_data=None):
             raise GraphQLError("Could not create share registry.")
 
 
+class OrganisationIdInput(graphene.InputObjectType):
+    id = graphene.ID(required=True)
+
+
+class AsicUpdate(graphene.ObjectType):
+    organisation = graphene.Field(Organisation)
+    user = graphene.Field(User)
```
cballantyne: I can return any field from user model here like password (granted it is hashed). I think we should just return the userId only. So not have a user graphene type. Unless there is reason to want to return other parts of user model.

ejjick:
Changes are in @cballantyne

User pw was never exposed before. GQ was only able to query fields that were already provided in the request, i.e `organisation.id` and `user.id`. I've removed the `user` altogether from the response. Thanks for noticing though, removing `user` should make us to think twice about adding it in the future.

PR #100, opened 2018-10-11
Feat/bs eventlog

bsippe:


web/eventlog/models.py
```diff
@@ -0,0 +1,15 @@
+import uuid
+from django.db import models
+from model_utils.models import TimeStampedModel
+from user_accounts.models import User
+
+
+class Eventlog(TimeStampedModel):
```
cballantyne: I thought we said no modified column as this is write only. So should extend from TimeStampedModel. Instead use models.Model.

web/eventlog/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from . import views
+
+urlpatterns = [
+    path('', views.index, name='index'),
+]
```
cballantyne: I don't this file or views.py below are necessary? There is no view for this module as it will have function that is used as events are picked off the queue and inserted into the table?

web/run_tests.sh
```diff
@@ -6,6 +6,7 @@ pip install -r requirements_test.txt
 export DJANGO_SECRET_KEY=1234abcd
 export DJANGO_SETTINGS_MODULE=web.settings.test
 flake8 --exclude=migrations && pytest \
+        --cov=eventlog \
```
cballantyne: I added this since eventlog was missing from coverage report. Can you please add a test to insert the model to the db. This should get the coverage report for this PR passing.

web/web/urls.py
```diff
@@ -33,6 +33,7 @@
     path('', include('user_accounts.urls')),
     path('admin/', admin.site.urls),
     path('kyc/', include('kyc.urls')),
+    path('eventlog/', include('eventlog.urls')),
```
cballantyne: I don't think this is needed as there are no views for this. Events will be processed as they come off the queue.

dpnova:
Does this PR need to have the queue + service stuff in it?

web/eventlog/models.py
bsippe: Agree - I've amended model type to models.Model to remove modified date field

web/eventlog/models.py
```diff
@@ -0,0 +1,15 @@
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    created = models.DateTimeField()
+    namespace = models.CharField(max_length=50)
+    payload = models.TextField()
+    user = models.ForeignKey(
+        User, db_index=True, on_delete=models.SET_NULL, null=True, blank=True
+    )
+    traceid = models.CharField(max_length=100)
```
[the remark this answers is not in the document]
bsippe: Yes agree - changed field name

cballantyne:
> Does this PR need to have the queue + service stuff in it?

We can do that in separate PR right after this one. Keep them smaller and get flow going.

web/run_tests.sh
troowala: https://github.com/enhancedsociety/platform/commit/4e63bcdb15c13806ad53f8a44166869866cd3da7

web/eventlog/models.py
troowala: https://github.com/enhancedsociety/platform/commit/5744b3a0c1ec7f3e70e1b904cab24e84188e321f

web/eventlog/admin.py
```diff
@@ -0,0 +1,34 @@
+from django.contrib import admin
+from eventlog.models import Eventlog
+
+
+class EventlogAdmin(admin.ModelAdmin):
+
+    list_display = ['user', 'namespace', 'payload']
+    readonly_fields = ['id', 'created', 'namespace', 'payload', 'user', 'trace_id']  # noqa: E501
```
dpnova: Can just spread these over multiple lines:

```
readonly_fields = [
    'f1', 'f2',
    'f3', 'f4',
]
```

cballantyne:
Code cov on patch only low due to admin config. Approving.

PR #102, opened 2018-10-12
Security on asic update mutation

cballantyne:
Can only be executed by user linked to officeholder for organisation.

### Pull Request check-list


web/web/schema.py
```diff
@@ -264,22 +264,41 @@ def mutate(root, info, asic_update_input=None):
...
 
 
+class OfficeholderUtils:
+    @staticmethod
+    def is_officeholder(organisation_id, user_id):
+        officeholders = OfficeholderModel.objects.filter(
+            organisation__id=organisation_id,
+            user__id=user_id
+        )
+        if len(officeholders) == 1:
```
dpnova: This will iterate the whole query from the db, deserialising every object. Better to use  ` return Foo.filter(...).exists()`

web/web/schema.py
```diff
@@ -264,22 +264,41 @@ def mutate(root, info, asic_update_input=None):
...
+                    info.context.user.id
+                ):
+                    raise Exception("Not authorised")
+                else:
+                    organisationModel = OrganisationModel(
+                        id=asic_update_input.organisation_id)
+                    asicUpdateModel = AsicUpdateModel(
+                        organisation=organisationModel,
+                        user=info.context.user)
+                    asicUpdateModel.save()
```
dpnova: I'd like to know for sure whether this saves a new org each time? Probably not but haven't used this pattern in my own code before

cballantyne: It does create a new record each time. I thought that is what it is supposed to do. Is that wrong?

cballantyne: Ah okay get you now. It does not create a new org. It only creates new asci update records.

cballantyne:
Merging as approved by @dpnova

PR #104, opened 2018-10-16
Feat/b org add acn

bsippe:


web/web/tests/test_graphene.py
```diff
@@ -43,7 +43,7 @@ def test_CreateShareRegistry_mutate_success(self):
         mutation myFirstMutation {
         createShareRegistry(shareRegistryData:
         {
-        organisation: {name: "Acme Pty Ltd"},
+        organisation: {name: "Acme Pty Ltd", acnAbn: "123456789"},
```
cballantyne: assert that the acnAbn was actually saved. 
Just like further down it does for the name:
self.assertEqual(
            graphql_organisation['name'],
            organisation.name
        )

bsippe: I've added this.

PR #105, opened 2018-10-16
mutation to modify shareholder email and address

cballantyne:
Can only be run by user linked to shareholder being updated or by user linked to officeholder for org shareholder belongs to.

### Pull Request check-list


web/web/schema.py
```diff
@@ -298,11 +340,22 @@ def is_officeholder(organisation_id, user_id):
             return False
 
 
+class ShareholderUtils:
+    @staticmethod
+    def is_shareholder(shareholder, user_id):
```
ejjick: This one is a bit confusing. While `is_officeholder` means officeholder of organisation, this one means shareholder entry is owned by user... Should we reserve `is_shareholder` for when it applies to an organisation and rename this one?

cballantyne: Good call. Trying to think what could be renamed to though. is_linked_to_shareholder?

ejjick: Haha yeah something like that, what I came up with was as verbose... `is_shareholder_owner`

cballantyne: going with - is_user_linked_to_shareholder. have refactored and cleaned up some prints in one of the tests and pushed.

PR #107, opened 2018-10-17
Feat/tr eventqueue

troowala:
### Pull Request check-list

eventlog
Web/schema.py

### Description of change
Event queue Feature for currently defined graphql mutations
CreateShareRegistry , CreateShareType, AddShareholder, TriggerAsicUpdate


web/web/schema.py
```diff
@@ -133,6 +134,7 @@ class Arguments:
     def mutate(root, info, add_shareholder_data=None):
         try:
             shareholderModels = []
+            log.apply_async(args=['shareholder.add', add_shareholder_data, info.context.user.id])  # noqa: E501
```
cballantyne: Pass published through as a param here.

web/web/tests/test_graphene.py
```diff
@@ -207,12 +219,18 @@ def test_CreateShareRegistry_mutate_error(self, organistation_save):
         }
   ''', context=context_mock)  # noqa
 
+        print(executed)
```
cballantyne: Remove print.

web/web/tests/test_graphene.py
```diff
@@ -150,11 +156,17 @@ def test_CreateShareRegistry_mutate_success(self):
...
-    def test_CreateShareRegistry_mutate_error(self, organistation_save):
+    def test_CreateShareRegistry_mutate_error(
+        self,
+        organistation_save,
+        eventlog_tasks_log_apply_async_mock
+    ):
         organistation_save.side_effect = Exception('Exception from mock.')
         context_mock = Mock()
-        # context_mock.user = user
+        # context_mock.user = '65888f1a-faa4-4385-8973-1d1eabe22836'
```
cballantyne: Remove these comments since they are not used.

web/eventlog/models.py
```diff
@@ -4,9 +4,9 @@
 
 class Eventlog(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    created = models.DateTimeField()
-    published = models.DateTimeField()
+    created = models.DateTimeField()  # datetime created in db
```
dpnova: This should have a default..

web/eventlog/tasks.py
```diff
@@ -0,0 +1,18 @@
...
+from eventlog.models import Eventlog
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task
+def log(namespace, payload, user, published):
+    logger.debug(f'Saving Eventlog : {namespace} by {user}')
+    Eventlog(
+        created=timezone.now(),  # datetime created in db
```
dpnova: This should be done by the default value on the field. https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.DateField.auto_now_add

web/eventlog/models.py
dpnova: https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.DateField.auto_now_add

web/web/schema.py
```diff
@@ -133,6 +135,7 @@ class Arguments:
     def mutate(root, info, add_shareholder_data=None):
         try:
             shareholderModels = []
+            log.apply_async(args=['shareholder.add', add_shareholder_data, info.context.user.id, timezone.now()])  # noqa: E501
```
dpnova: Should we prefix these with `es.` ? Just in case we end up with other namespaces in the event log?

web/eventlog/tests/test_eventlog.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.assertEqual(
+            self.uuid,
+            str(p.trace_id)
+        )
+
+        p.delete()
+
+    def test_log_function(self):
+        log(self.namespace, self.payload, self.uuid, self.now)
```
ejjick: Not sure if we want to do this. It will call the "write to db task" function directly, without adding anything to the queue... see `kyc/test_kyc_views.py` line 132 for an example of how to mock the queue.

web/web/schema.py
```diff
@@ -270,6 +278,8 @@ def mutate(root, info, asic_update_input=None):
                 ):
                     raise Exception("Not authorised")
                 else:
+                    log.apply_async(args=['asicupdate.trigger', asic_update_input, info.context.user.id, timezone.now()])  # noqa: E501
```
dpnova: Also thinking it might be good to tie the namespace to the structure of the app a bit more?

web/web/schema.py
```diff
@@ -197,7 +202,10 @@ class Arguments:
 
     @staticmethod
     def mutate(root, info, share_registry_data=None):
+
         try:
```
dpnova: I'd like us to break this try accept pattern at some point... perhaps a refactor after this @cballantyne ?

web/web/schema.py
troowala: can we at some point have a quick chat and decide upon the nomenclature for namespaces ? @cballantyne @ejjick 

also since this is smart equity if we are prefixing it then would it be se. and not es. ?

cballantyne: yeah something like se. would be a good prefix. (smartequity)

web/web/schema.py
troowala: yep , agreed, lets figure out a naming convention for these things in our next engineering discussion.

web/web/schema.py
cballantyne: Not sure what you mean. I thought that was how mutate methods were done. Happy to be pointed in the right direction if I have set it up wrong. 

Lots of refactoring to do when the dust settles on this though. case fixing, get it all out of one file, connecting share types properly to shareholdings.

web/eventlog/models.py
```diff
@@ -4,9 +4,9 @@
...
+    published = models.DateTimeField()  # datetime pushed onto queue
     namespace = models.TextField()
     payload = models.TextField()
     user = models.UUIDField()
-    trace_id = models.TextField()
+    trace_id = models.TextField(null=True)
```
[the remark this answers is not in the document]
troowala: ok

web/eventlog/models.py
troowala: ok

web/eventlog/tests/test_eventlog.py
troowala: ignoring this comment as the queues are being tested in test_graphene.py , this is for the actual log function test to get better coverage. if i patch & mock this here then the log functionality is not tested.

PR #110, opened 2018-10-18
restrict queries to officeholders

ejjick:
### Pull Request check-list


web/web/schema.py
```diff
@@ -491,10 +506,14 @@ def resolve_sharetypes(self, info, organisation_id=None):
...
+        elif not OfficeholderUtils.is_officeholder(
+            organisation_id,
+            info.context.user.id
+        ):
+            raise Exception("Not authorised")
         else:
             return SharetypeModel.objects.filter(
-                organisation__id=organisation_id,
-                organisation__created_by=info.context.user)
+                organisation__id=organisation_id)
```
cballantyne: This is great the the filter no longer includes created_by. Can you please add this in for shareholders, officeholders and organisation resolve methods. As having created_by in these doesn't make sense anymore now that we have security that user needs to be an office holder.

cballantyne: I have fixed this up.

ejjick: Thanks mate. Missed that one. Removed it from the others.

PR #111, opened 2018-10-18
freeze shareholder

troowala:
### Pull Request check-list

SE 

### Description of change
Freeze shareholder functionality under Admin area for ASIC compliance.


web/se/models.py
```diff
@@ -29,6 +29,11 @@ class Shareholder(TimeStampedModel):
     sharetype = models.TextField()
     payment_status = models.TextField()
     email = models.EmailField(null=True, blank=True)
+    eth_address = models.CharField(
+        null=True, blank=True,
+        max_length=42, db_index=True)
+    onchain = models.BooleanField(default=False)
```
cballantyne: I am unsure about putting onchain against a shareholder. An organisation shares are either onchain or they are not? It is all or nothing. Not sure how this would affect shareholders without eth_addresses. Would need to discuss this.

web/se/admin.py
```diff
@@ -20,7 +21,43 @@ class OfficeholderAdmin(admin.ModelAdmin):
 
 class ShareholderAdmin(admin.ModelAdmin):
 
-    list_display = ['organisation', 'name']
+    def freeze_shareholder(modeladmin, request, queryset):
```
cballantyne: It would be nice to have a test for these functions. You would have to Mock out TokenManager.

web/se/admin.py
```diff
@@ -20,7 +21,43 @@ class OfficeholderAdmin(admin.ModelAdmin):
...
+        for query in queryset:
+                log('se.shareholder.freeze', str(query.id), request.user.id, timezone.now())  # noqa : E501
```
cballantyne: I thought we were only logging events that came through graphql as mutations. @dpnova  - is that right?

web/se/admin.py
troowala: Done. https://github.com/enhancedsociety/platform/pull/111/commits/305a33a04a11b43b1596b2abb0635602c7d709a3

web/se/admin.py
troowala: as per discussion , since this is needed for compliance we will log it. Payload format from admin area functions TBD at a later date. Currently capture the event and the shareholder id & txid (if applicable)

PR #118, opened 2018-10-23
initial commit.

cballantyne:
### Pull Request check-list


web/web/tests/test_graphene.py
```diff
@@ -108,7 +108,8 @@ def test_CreateShareRegistry_mutate_success(
             name:"Jane Doe", 
             address:"22 Elsewhere St, Burleigh Heads, 4220, QLD, AUSTRALIA", 
             role: "CEO", 
-            birthdate:"1978-12-20"
+            birthdate:"1978-12-20",
+            email: "jane.doe@example.com"
```
ejjick: We could start using `Faker` to create all these dummy values

PR #120, opened 2018-10-24
Feat/tr test contracts

troowala:
### Pull Request check-list

SE, Dockerfile

### Description of change
Move solc dependency to be downloaded at docker-compose build instead of docker-compose up and retry codecov tests.


troowala:
this will ensure solc gets downloaded prior at the initial docker container build for web , as some docker caching issues result in it being re-downloaded every so often otherwise.

also tests have been created so they work soon as skip is removed once ganache connectivity can be established.

dpnova:
Unit tests shouldn't depend on ganache. Let's just mock the things that are outside our codebase

cballantyne:
These are going to be integration tests that we will be running later using docker compose. 
When all the dust settles on latest bunch of PRs I will be having a runner for unit tests and another one for integration tests.

PR #125, opened 2018-10-25
refactor shareholder sharetype association

troowala:
### Pull Request check-list

SE

### Description of change
efactor createShareRegistry to correctly create and link the share types.


web/se/models.py
```diff
@@ -68,8 +48,33 @@ class Sharetype(TimeStampedModel):
...
     contract_address = models.CharField(
         max_length=42,
         help_text="Leave Blank to create new token if Onchain is selected",
         null=True, blank=True)
+
+    def __str__(self):
+        return self.symbol
+
+
+class Shareholder(TimeStampedModel):
```
cballantyne: Any chance we could have this in the place it was originally in the file at line 31. It shows a lot of noise in the diffs that is all. Makes it harder to see what actually changed. Also might be the cause of the merge conflict on se/schema.py.

troowala: as discussed , this had to be moved below Sharetype as it is a dependency for Shareholder

web/se/models.py
```diff
@@ -68,8 +48,33 @@ class Sharetype(TimeStampedModel):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    organisation = models.ForeignKey(
+        Organisation, on_delete=models.CASCADE, db_index=True)
+    user = models.ManyToManyField(User, db_index=True, blank=True)
+    name = models.TextField()
+    address = models.TextField()
+    shares = models.IntegerField()
+    sharetype = models.ForeignKey(
+        Sharetype, db_index=True, on_delete=models.DO_NOTHING,
+        null=True, blank=True)
```
cballantyne: I don't think we should allow nulls here. 
And when adding a shareholder either by addShareholder or createShareRegistry it could check whether the share symbol exists and return a message like 'Sharetype does not exist' or something similar. 
Otherwise if we end up with shareholders with no sharetype that would cause a bunch of other problems.

troowala: sorry , do you mean shouldnt allow nulls ?

troowala: and also not allow blanks either then im guessing

web/se/schema.py
```diff
@@ -342,14 +338,30 @@ def mutate(root, info, share_registry_data=None):
...
             for shareholder in share_registry_data.shareholders:
                 shareholderModel = ShareholderModel(
                     id=uuid.uuid4(),
                     organisation=organisationModel,
                     name=shareholder.name,
                     address=shareholder.address,
-                    sharetype=shareholder.share_type,
+                    sharetype_id=SharetypeModel.objects.filter(
+                        symbol=shareholder.share_type,
+                        organisation_id=organisationModel).first().id,
```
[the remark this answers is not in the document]
troowala: as per discussion , we will stick with the current way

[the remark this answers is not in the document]
cballantyne: Okay can leave this for now. Will revisit later for performance optimise.

web/se/models.py
dpnova: You can use `'Sharetype'` instead of `Sharetype` to avoid the dep ordering issue

web/se/schema.py
```diff
@@ -380,6 +392,7 @@ def mutate(root, info, share_registry_data=None):
 
             share_registry = ShareRegistry(
                 organisation=organisationModel,
```
dpnova: Can we bin the camel case on all these?

web/se/signals.py
```diff
@@ -0,0 +1,17 @@
...
+from se.models import Sharetype
+from se.contracts import TokenManager
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+# Below function is called to create a token if doing a Sharetype.save() with onchain = true   # noqa : E501
+@receiver(pre_save, sender=Sharetype)
+def my_callback(sender, instance, *args, **kwargs):
```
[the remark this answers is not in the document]
dpnova: Yeah..

web/se/schema.py
troowala: yep , my understanding is B is working on a PR which uniforms the casing all over. that will sort this out

PR #129, opened 2018-10-30
Graphql Schema Documentation

troowala:
Documentation configuration & content for graphql mutations and queries.
Also ensures graphql tests leverage documentation content at run time where applicable.

SE , TSE , TRADING , USER_PROFILE


web/se/schemadoc.py
```diff
@@ -0,0 +1,343 @@
...
+    query Example {organisation
+        (organisationId: "86318182-005b-427a-a3af-be07230a5e58") {
+                id
+                name
+            }
+        }
+    """
+
+    resolve_sharetypes = """
+    # This queryies the list of attributes for a specified Sharetype.
```
cballantyne: spelling - queries.

cballantyne:
How can we generate the html file with the graphql docs?
Any chance this could be included as part of the docs generation for the platform? If not just the method to generate the graphql html docs file you attached in the ticket is fine.

cballantyne:
I love the way the docs are part of the unit tests. It means the docs should never have wrong examples ... which is a pain for devs and happens so often.

PR #131, opened 2018-10-31
Add officeholder req fields.

troowala:
### Pull Request check-list

SE

### Description of change
Mandate birthstate , birthcity , birthcountry fields for AUS officeholders


web/se/schema.py
```diff
@@ -272,29 +276,35 @@ def mutate(root, info, add_officeholder_data=None):
...
+                        email=add_officeholder_data.officeholder.email
+                    )
+                    officeholder_model.save()
             else:
-                raise Exception("Not authorised")
+                raise SmartEquityException("Not authorised")
 
             return AddOfficeholder(officeholder=officeholder_model)
 
+        except SmartEquityException as e:
```
cballantyne: @dpnova @troowala - I have tweaked this so we throw SmartEquityException. Previously if unexpected error occurred we could have bled out internal details to API users. This way we give the extra message if we know it was an exception thrown by us with a particular message. Otherwise it just gives the generic message. 
Does this seem okay?

troowala: seems better than okay ! this is neat ... i get it now.

troowala:
can i merge this now ?

PR #133, opened 2018-11-02
Feat/cb integration tests

cballantyne:
Separate integration tests identified by files in test prefixed with 'it' for integration test.

Codepipeline has IntegrationTest stage which runs integration tests using docker compose and the actual that was created from the build stage.

### Pull Request check-list


.travis.yml
```diff
@@ -4,22 +4,26 @@ python:
...
 
-addons:
-        hosts:
-                - ganache
-
-before_install:
-        - npm install -g ganache-cli
-
 install:
+        # These next 2 steps should not be needed but travis doesn't seem to handle shutting down postgres from previous runs.
```
cballantyne: I think we don't want docker compose running here because of how long it takes to grab images for postgres etc and then build the image for our platform?
At least not yet until we could figure out if caching possible.

PR #135, opened 2018-11-02
Refactor Create Sharetype onchain to be done using event queues

troowala:
### Description of change
"Refactor Create Sharetype onchain to be done using event queues"

SE

### Pull Request check-list


web/se/admin.py
```diff
@@ -38,7 +37,7 @@ def unfreeze_shareholder(modeladmin, request, queryset):
         for query in queryset:
                 log('se.shareholder.unfreeze', str(query.id), request.user.id, timezone.now())  # noqa : E501
                 if query.eth_address:
-                    sharetype_contract_address = Sharetype.objects.filter(name__iexact=query.sharetype).first().contract_address  # noqa : E501
+                    sharetype_contract_address = Sharetype.objects.filter(name__iexact=query.sharetype.name).first().contract_address  # noqa : E501
```
cballantyne: Why are we querying on the name instead of the ID of the sharetype?

web/se/admin.py
```diff
@@ -66,8 +65,15 @@ class AsicUpdateAdmin(admin.ModelAdmin):
 
 
 class SharetypeAdmin(admin.ModelAdmin):
+
     list_display = ['organisation', 'name', 'symbol',
                     'quantity', 'onchain', 'contract_address']
 
+    fieldsets = [
```
cballantyne: If we don't have onchain functionality in admin why are the methods defined in admin.py?

troowala: "Creation of onchain token" not all onchain functionality.

The functionality in admin area is for freeze/unfreeze shareholder.

PR #136, opened 2018-11-02
fixes eth_address login related issues currently on ES website.

troowala:
### Pull Request check-list


web/user_accounts/views.py
```diff
@@ -111,10 +113,17 @@ def setup_ethereum_address(request):
...
                 'registration/setup_address.html',
                 {**ctx, 'eth_form': form})
-        addr_obj = EthereumAccount(eth_address=addr, user=request.user)
-        addr_obj.save()
-        # TODO if `next` isn't provided, where do we redirect to? root?
-        return HttpResponseRedirect(request.POST.get('next', '/'))
+
+        EthereumAccount(eth_address=addr, user=request.user).save()
+
+        if Profile.objects.filter(user=request.user) is None:
```
cballantyne: I think Profile.objects.filter(user=request.user).exists() is better to use in this case.

PR #138, opened 2018-11-12
getedge API Example

troowala:
Allows writes to ASIC using the getedge API.

This example shows how to leverage getedge API form for adding a new officeholder by setting, 

onedge : true 

in GraphQL mutation and the entry will also be written to ASIC via edge.

ToDo:

 - Make integration use eventqueues
- Tests
- Still some accesses are pending from getedge / Asic for the entire workflow to be complete , for example querying back and confirming all good and requiring registered agent access for companies we need to update for.


web/se/edgeapi.py
```diff
@@ -0,0 +1,88 @@
...
+class forms():
+
+    def __init__(self):
+
+        self.headers = {
+            'Content-Type': 'application/json',
+            'X-Auth-Edge': settings.GETEDGE_API_KEY
+            }
+
+    def recursive_urlencode(d):
```
cballantyne: Some more comments please why this is necessary and some existing built in lib cannot be used.

web/se/edgeapi.py
```diff
@@ -0,0 +1,88 @@
...
+
+        return '&'.join(recursion(d))
+
+    def RA71(self, **kwargs):
+
+        """
+            Lodge an RA71 Form for Adding/Updating Office Details
+            Returns a Response ID.
+        """
+        url = 'https://getedge.com.au/api/v1/lodge71'
```
cballantyne: Do they have a test api gateway? Should this be deploy config?

web/se/edgeapi.py
```diff
@@ -0,0 +1,88 @@
...
+
+        fields = {
+            "identifier": datetime.datetime.now().strftime("%s"),
```
cballantyne: I prefer to pass any datetime into function. Easier to test. Can this be a param please? Same for date_signed below.

web/se/edgeapi.py
```diff
@@ -0,0 +1,88 @@
...
+                },
+            "company_id": {
+                "company_name": kwargs.get('company_name'),
+                "acn": kwargs.get('acn'),
+                "type": "DATA"
+                }
+            }
+
+        data = forms.recursive_urlencode(fields)
+        print(data)
```
cballantyne: Remove print. Or replace with log statement.

web/se/schema.py
```diff
@@ -146,6 +147,7 @@ class UpdateOfficeholderData(graphene.InputObjectType):
 class AddOfficeholderData(graphene.InputObjectType):
     officeholder = graphene.InputField(OfficeholderInput)
     organisation_id = graphene.String()
+    onedge = graphene.Boolean(required=False, default_value=False)
```
cballantyne: I don't think this should be passed with mutation call. It should either be deploy setting or a per org. Chat tomorrow about this with everyone.

web/se/edgeapi.py
troowala: unfortunately not for now :( , they will provide us one later when we have our own test creds from ASIC

web/se/schema.py
```diff
@@ -295,6 +297,25 @@ def mutate(root, info, add_officeholder_data=None):
                         email=add_officeholder_data.officeholder.email
                     )
                     officeholder_model.save()
+
+                    # If onedge is true then update using GetEdge API
+
+                    if add_officeholder_data.onedge:
```
cballantyne: Not sure how test coverage has increased with this PR. 
This maybe needs to be split out into separate method where officeholder data is mapped to data used by getedge form. And that could have a test.
Then getedge could be mocked and assert that it is called.
Can discuss this tomorrow after standup as well.

web/se/edgeapi.py
troowala: yep that was just for testing. it will be gone ...

web/se/schema.py
troowala: yep we can figure the best way out for this , for now to demonstrate the integration this seemed like a simple way

web/se/tests/test_edgeapi.py
```diff
@@ -0,0 +1,56 @@
...
+
+        self.assertEqual(
+            'DOE',
+            form.fields['signing_officer']['family_name']
+        )
+
+        self.assertEqual(
+            'JOHN',
+            form.fields['signing_officer']['given_name1']
+        )
```
cballantyne: Please assert against more of the fields. This is only doing signing_officer. There are lots more like officers, company_id.

web/infrastructure/deploy/main.tf
```diff
@@ -500,6 +500,7 @@ data "template_file" "task_definition" {
     app_resource_prefix = "${local.name_prefix}"
     app_uploads_storage_bucket_name = "${local.name_prefix}app-uploads.enhancedsociety.com"
     app_sqs_name = "${local.name_prefix}app-enhancedsociety-sqs"
+    getedge_api_key = "${var.getedge_api_key}"
```
cballantyne: Back this change out please. task definition env vars have changed since this branch was started.

web/infrastructure/deploy/task-definition.json
```diff
@@ -19,7 +19,8 @@
             {"name": "APP_ENVIRONMENT", "value": "${app_environment}"},
             {"name": "APP_RESOURCE_PREFIX", "value": "${app_resource_prefix}"},
             {"name": "APP_UPLOADS_STORAGE_BUCKET_NAME", "value": "${app_uploads_storage_bucket_name}"},
-            {"name": "APP_SQS_NAME", "value": "${app_sqs_name}"}
+            {"name": "APP_SQS_NAME", "value": "${app_sqs_name}"},
+            {"name": "GETEDGE_API_KEY", "value": "${getedge_api_key}"},
```
cballantyne: Back this change out. Task definition env vars have changed.

web/infrastructure/deploy/variables.tf
```diff
@@ -32,3 +32,82 @@ variable "git_commit" {
   description = "The git commit to uniquely ident."
 }
 
+variable "postgres_user" {
```
cballantyne: Do a merge with master which should remove all of these additions. Task definition env vars are handled differently now via infrastructure/config/.

web/se/edgeapi.py
```diff
@@ -0,0 +1,186 @@
...
+
+        organisation_name = str(
+            OrganisationModel.objects.filter(
+             id=org_id).first().name).upper()
+
+        organisation_acn = str(
+            OrganisationModel.objects.filter(
+             id=org_id).first().acn_abn).replace(" ", "")
+
+        officername = str.split(
```
cballantyne: Can these vars be a bit more readable please. Like officer_name, signer_firstname, etc.

web/se/edgeapi.py
```diff
@@ -0,0 +1,186 @@
...
+                        "number_of_pages":  "1",
+                        "text":  "APPOINTING OFFICEHOLDER"
+                }
+            }
+
+        return fields
+
+    def submit(self, user_id):
+        data = json.dumps(self.fields)
+        print(data)
```
cballantyne: Remove print.

web/se/schema.py
```diff
@@ -301,6 +303,16 @@ def mutate(root, info, add_officeholder_data=None):
                         email=add_officeholder_data.officeholder.email
                     )
                     officeholder_model.save()
+
+                    # if onedge is true
+                    if add_officeholder_data.onedge:
+
+                        # Lodge form 484 for adding officeholder via event queue  # noqa: E501
+                        lodgeaddofficeholder.s(
```
cballantyne: A test to assert that the task is called when onedge true would be good.

troowala: lets chat about the best way to test a celery task is being created upon a param being passed without actually following on the execution or if something is mocked we verify it is still being called. any existing tests we have that might be similar ?

cballantyne: Can leave this test for now because when onedge is called will come from org settings later on instead of being passed as value in graphql mutation.

PR #143, opened 2018-11-15
Feat/cb deploy get secrets

cballantyne:
config folder holds terraform for creating parameter store records.

task definition runserver_prod.sh tweaked to pull parameter store records and set them as environment variables.

Removes need to set environment variables via task definition which are viewable in console and make it harder for continuous deployment since the deploy step needs to know about the variable values.

### Pull Request check-list


web/infrastructure/config/main.tf
```diff
@@ -0,0 +1,243 @@
...
+resource "aws_ssm_parameter" "mandrill_user" {
+  name        = "/${local.name_prefix}app/env/MANDRILL_USER"
+  type        = "SecureString"
+  value       = "${var.mandrill_user}"
+  key_id = "${aws_kms_key.web_key.id}"
+
+  tags = "${local.common_tags}"
+}
+
+resource "aws_ssm_parameter" "madnrill_api_key" {
```
dpnova: typo?

cballantyne: Yeah.Fixed that.

PR #146, opened 2018-11-27
Added quantity_unissued, amount_paid, amount_outstanding to sharetype…

cballantyne:
ShareType has new fields quantity_unissued, amount_paid and amount_outstanding.

These are carried through to the mutations for createShareType and createShareRegistry.
Also to the query sharetypes.

### Pull Request check-list


web/web/schema.py
```diff
@@ -12,6 +12,7 @@
 class Mutation(
         se.schema.Mutation,
         organisations.schema.Mutation,
+        user_accounts.schema.Mutation,
```
cballantyne: This is intentional. Fixing bug in master.

PR #147, opened 2018-11-28
Feat/getedge addshareholder

troowala:
Enables AddShareHolder ASIC/EDGE Integration.

SE

### Description of change
Enables AddShareHolder ASIC/EDGE Integration.


web/se/tests/test_edgeapi.py
```diff
@@ -97,3 +117,98 @@ def test_addofficeholder_build_data(self):
              '4221',
              form.fields['officers'][0]['address']['postcode']
         )
+
+    def test_shareholder_build_data(self):
```
cballantyne: Name should have add in it, ie test_addshareholder_build_data.

troowala: thanks , will address in the upcoming PR with more tests. hope that is ok.

PR #148, opened 2018-11-28
Update schema.py

troowala:
ASIC requires share names to be PREFERENCE or ORDINARY , etc , cannot have org names in them , 

Also when tokens are create on chain they should be 0 , then when they are issued they can be minted to the appropriate holder. This is again an ASIC requirements that any shares issued should be issued to an entity and cant be blank.

hence these 2 minor changes are needed to maintain compatibility with ASIC.

### Pull Request check-list


cballantyne:
Maybe later.

PR #152, opened 2018-11-30
Fix/asic onchain compat

troowala:
- Maximum character limits addressed
- Better tests
- Tokens are not minted by default upon creating share type but instead now support calling mint explicitly only when issuing to a shareholder.

### Pull Request check-list

### Description of change
Tweaks to ensure our onchain functionality complies with edge/asic.


web/se/schema.py
```diff
@@ -376,13 +376,24 @@ def mutate(root, info, create_sharetype_data=None):
 
         if create_sharetype_data.sharetype.onchain:
             if not create_sharetype_data.sharetype.contract_address:
+
+                organisation_name = str(
+                        OrganisationModel.objects.filter(
+                            id=create_sharetype_data.organisation_id)
+                        .first().name).upper().replace("PTY LTD", "")
```
cballantyne: I see this an interim solution to ensuring token name is unique across orgs. But okay for now.

PR #154, opened 2018-12-05
Feat/edit shareholder asic

troowala:
### Pull Request check-list
This allows for updating Shareholder Addresses and their Holdings in ASIC.

SE

### Description of change
This allows for updating Shareholder Addresses in ASIC.


web/se/schema.py
```diff
@@ -187,11 +188,64 @@ def mutate(root, info, update_shareholder_data=None):
...
+                update_shareholder_data.shareholder.address
+            shareholder_model.save()
+
+        # Check for change in Shareholder Email
+        if (e.email != update_shareholder_data.shareholder.email):
+
+            # Update Shareholder Email
+            shareholder_model.email = \
+                update_shareholder_data.shareholder.email
+            shareholder_model.save()
```
cballantyne: If email, address and shares are updated there will be 3 separate db writes. Is this intended?
Cleaner as more separate. I guess this scenario won't happen much so won't be a performance issue.

web/se/tasks.py
```diff
@@ -51,3 +52,35 @@ def lodgeaddshareholder(payload, user_id, now, update_id):
         id=update_id).update(
         responseid=responseid,
         status='LODGED')
+
+
+@shared_task
+def lodgeupdateshareholder_address(payload, user_id, now, update_id):
```
cballantyne: I cannot see how this is being triggered in this PR. Is it part of this PR?

web/se/schema.py
troowala: yep , we did three separate writes, because this allows updating of shareholder address and shares separately incase the user makes a mistake with one the other should not fail. 

also the number of criterias to consider if both address is changds and so is share quantity when done together can cause some issues such as the member is not found when a different address is specified , so you first got to update shares against the existing member record with the existing address and then you got to update the address , or vice versa but you try n do both it gets pretty convoluted.

as for email we dont update that with asic as asic doesnt require that.

web/se/tests/test_schema.py
```diff
@@ -304,6 +304,16 @@ def test_add_shareholder_mutate_success(
...
         )
 
+        # Ensure a PENDING Entry was written to ASIC Update Queue
+        self.assertEqual(
+            'PENDING',
+            AsicUpdate.objects.filter(
+                organisation_id='b280926e-8bae-4481-9d2c-d6a2d969400b',
+                user=user,
+                task='lodgeaddshareholder'
+                ).first().status
```
cballantyne: There should only be 1 record in AsicUpdate table? So can this be asserted please.

web/se/tests/test_schema.py
```diff
@@ -959,6 +923,28 @@ def test_updateshareholder_mutate_success(
...
+        # Ensure PENDING Entries were written to ASIC Update Queue
+
+        # For Address Update
+        self.assertEqual(
+            'PENDING',
+            AsicUpdate.objects.filter(
+                organisation=organisation,
+                user=user,
+                task='lodgeupdateshareholder_address'
+                ).first().status
```
cballantyne: There should only be 1 record in AsicUpdate table? So can this be asserted please.
To catch case where possibly another one was written for update_email when it should not have been.

troowala: addressed,

web/se/tests/test_schema.py
troowala: asserted now in the latest commit
https://github.com/enhancedsociety/platform/pull/154/commits/abf36b166c0a9ec482c20fac31dd96104eebbc4b

PR #155, opened 2018-12-05
Asic org deets

dpnova:
### Pull Request check-list


web/organisations/models.py
```diff
@@ -6,6 +6,16 @@
 from user_accounts.models import User
 
 
+class Address(TimeStampedModel):
+    care_of = models.CharField(max_length=200)
+    line2 = models.CharField(max_length=200)
+    street = models.CharField(max_length=200)
```
cballantyne: Should street be line1?

web/organisations/models.py
```diff
@@ -6,6 +6,16 @@
 from user_accounts.models import User
 
 
+class Address(TimeStampedModel):
```
cballantyne: Should this have a uuid ID field? Otherwise it makes it auto inc by default.

web/organisations/schema.py
```diff
@@ -17,11 +85,9 @@ class Meta:
 
 class CreateOrganisationInput(graphene.InputObjectType):
     name = graphene.String(required=True)
-    acn_abn = graphene.String(required=True)
-    website = graphene.String(required=True)
-    contact_email = graphene.String(required=True)
-    contact_phone = graphene.String(required=True)
-    contact_role = graphene.String(required=True)
+    acn_abn = graphene.String(required=False)
```
cballantyne: acn_abn can be removed as well since it is contained under asic_details?

web/organisations/tests/test_org_schema.py
```diff
@@ -15,13 +22,63 @@ def setUp(self):
         self.resolve_signed_in_mock.context = self.context_signed_in_mock
         self.context_signed_in_mock.user = UserFactory.create()
 
+    @mock.patch('organisations.schema.create_org_from_inputs')
+    def test_mutate_w_asic(self, create_org_from_inputs):
```
cballantyne: better name for test. This is not testing with asic? It is testing that partner record is created?

web/organisations/tests/test_org_schema.py
```diff
@@ -30,11 +87,19 @@ def test_mutate_partner(self):
         )
         self.assertEqual(PartnerOrganisation.objects.count(), 1)
 
+    def test_address_from_input(self):
+        from organisations.schema import address_from_input
+        ainput = AddressInput()
+        ainput.care_of = "Someone"
+        addr = address_from_input(ainput)
+        self.assertEqual(addr.care_of, "Someone")
```
cballantyne: Could we test all attributes like suburb, line2, etc.

web/organisations/tests/test_org_schema.py
```diff
@@ -15,13 +22,63 @@ def setUp(self):
...
+        )
+        data.asic_details.registration_date = "2018-04-05"
+        data.asic_details.next_review_date = "2018-04-05"
+        data.asic_details.name_start_date = "2018-04-05"
+        data.asic_details.registration_date = "2018-04-05"
+        data.asic_details.registered_address_start_date = "2018-04-05"
+        data.asic_details.contact_address_start_date = "2018-04-05"
+        data.asic_details.principle_place_of_business_address_start_date\
+            = "2018-04-05"
+        data.asic_details.registered_address = AddressInput()
```
cballantyne: Could we test with an address? The registered address should always be present anyway?

web/se/schema.py
```diff
@@ -420,18 +417,15 @@ def mutate(root, info, share_registry_data=None):
...
-            id=uuid.uuid4(),
-            name=share_registry_data.organisation.name,
-            created_by_id=info.context.user.id,
-            acn_abn=share_registry_data.organisation.acn_abn
-        )
+        organisation, asic_details = create_org_from_inputs(
+            share_registry_data.organisation,
+            info.context.user.id)
         organisation.save()
+        if asic_details:
```
cballantyne: The test for this mutation would be using the docs which were updated with asic_details. Could the test also be updated to make sure that the correct data is being saved and returned?

web/organisations/models.py
dpnova: Actually I think @Enhanced-Johan and @kimhansen agreed on something here?

web/organisations/models.py
dpnova: Yeah good point, I forget it appears in gql.

web/organisations/models.py
troowala: nope its fine. this is how ASIC form is
'registered_office': {
                # 'date_of_change': '',
                # 'address': {
                #     'care_of': '',
                #     'line2': '',
                #     'street': '',
                #     'locality': '',
                #     'state': '',
                #     'postcode': '',
                #     'country': ''
                # },

web/organisations/schema.py
troowala: is you remove acn/abn ... it will affect signer authorization for submitting forms.
it obtains the acn/abn details from here , however i do agree at some point acn_abn should be changed to separate acn and abn fields

dpnova: I didnt want to break backwards compat immediately here...

dpnova: The view is the org doesn't have any AUS specific stuff on it

web/organisations/models.py
cballantyne: This is fixed now.

web/se/schema.py
dpnova: Yeah I haven't added those graphql query tests...

dpnova: The current tests are unit tests - no db to return things from. I could check the params of "create" were called properly.


Do we need the graphql tests? I'm not sure I'll have time to add them today

dpnova:
I could probably add some graphql tests using the doc method we have.. I'm not sure I'll get time today @cballantyne

PR #157, opened 2018-12-12
Feat/shareholder amounts paid and middlename support

troowala:
### Pull Request check-list

Support price per share for shareholder entries
Support middlenames for shareholders

SE

### Description of change
Support price per share for shareholder entries
Support middlenames for shareholders


web/se/schema.py
```diff
@@ -218,6 +221,11 @@ def mutate(root, info, update_shareholder_data=None):
                 update_shareholder_data.shareholder.shares
             shareholder_model.save()
```
cballantyne: This could be removed since there is a save just after price_per_share is set.

web/se/schema.py
```diff
@@ -218,6 +221,11 @@ def mutate(root, info, update_shareholder_data=None):
...
 
+            # Update Shareholder Shares
```
cballantyne: I don't think we need this comment.

web/se/tests/test_edgeapi.py
```diff
@@ -143,7 +145,7 @@ def test_addofficeholder_build_data(self):
     def test_addshareholder_build_data(self):
 
         # Shareholder Details
-        name = 'John Doe'
+        name = 'Jane Doe Smith'
```
cballantyne: Why is shareholder name different than the user linked to the shareholder?

web/se/tests/test_edgeapi.py
```diff
@@ -194,7 +197,7 @@ def test_addshareholder_build_data(self):
         )
 
         self.assertEqual(
-             20,
+             5000,
```
cballantyne: I thought it should be 20 x 2.50 = 50. Why are there extra zeros?

web/se/tests/test_edgeapi.py
```diff
@@ -208,6 +211,11 @@ def test_addshareholder_build_data(self):
              form.fields['share_issue']['share_classes'][0]['total_number_in_issue']  # noqa : E501
         )
 
+        self.assertEqual(
+             2500,
```
cballantyne: Amount paid should be 2.50 x 10 = 25. Or I am not understanding it correctly?

web/se/tests/test_edgeapi.py
```diff
@@ -233,6 +241,36 @@ def test_addshareholder_build_data(self):
...
+             'JANE',
+             form.fields['members'][0]['holding_owners'][0]['member_name_person']['given_name1']  # noqa : E501
+        )
+
+        # Ensure Sharetype QTY total in DB is Increased
+        sharetype_1 = Sharetype.objects.filter(
+           id='beb6dd77-53b1-4f5a-9e43-da52c1eb8973').first()
+
+        self.assertEqual(
+             20,
```
cballantyne: How is sharetype quantity and amount_paid increased? I cannot see how this is done in the code.

web/se/tests/test_edgeapi.py
troowala: ASIC requires amounts in cents ...

web/se/tests/test_edgeapi.py
troowala: ASIC requires amounts in cents .

web/se/tests/test_edgeapi.py
troowala: edgeapi.py lines 789-794
       
# Update Share Type Overall Qty.
        SharetypeModel.objects.filter(
            organisation_id=org_id,
            symbol=share_type).update(
            quantity=total_shares_qty,
            amount_paid=total_amount_paid)

web/se/schema.py
troowala: done

web/se/schema.py
troowala: removed

web/se/tests/test_edgeapi.py
cballantyne: I am not going to be a scope creep. I just noticed that is already in the codebase. We will have to do a separate PR to correct this. Leave it as is for now.

troowala: i agree with you , the reason this is already in build data is because build data needs previous amounts and future amounts and once future amounts are calculated either it will call a separate function to write this info or it will require previous data values to be passed through , so yes im happy to refactor this to a better way but for now allowing it and making a note in asana under TODO

troowala:
Merging now , will deploy shortly with auto deploy PR coming through.

PR #161, opened 2018-12-19
Feat/cb org settings

cballantyne:
New table for storing which modules are enabled for an organisation. 
Currently there is only one - asic_update.

The organisationmodules will be returned for any queries that return an organisation like partnerOrganisations, shareholderOrganisations, officeholderOrganisations. This will be used by the frontend to disable the 'update asic' button.

If an organisation does not have asic update enabled a mutations like addShareholder, updateShareholder or addOfficeholder would not insert a record into the AsicUpdate table.

### Pull Request check-list


web/organisations/models.py
```diff
@@ -69,3 +69,14 @@ class ASICOrganisationDetails(TimeStampedModel):
     registered_address = models.ForeignKey(
         Address, on_delete=models.CASCADE, related_name="orgs_registered")
     registered_address_start_date = models.DateField(null=True)
```
ejjick: May need `blank=True` too, unless we want the field to be mandatory in forms.

cballantyne: This is for a model outside this PR - it was already there. So I am not going to change it in this PR. I think this field is required in forms so not sure why it allows null in db.

cballantyne: The model for this PR OrganisationModules has asic_update with default=false. It does not have blank=true but I think that is okay since this is only done via admin.

PR #162, opened 2018-12-20
"Check Status After Lodgement"

troowala:
### Pull Request check-list
This will query the status of asic lodgments every 2 minutes after lodgement and update the status column to sucessful or failed.

TO DO:
Find out reason why it failed.

SE

### Description of change
query the status of asic lodgments every 2 minutes after lodgement and update the status column to sucessful or failed.


web/se/edgeapi.py
```diff
@@ -65,7 +105,7 @@ def build_data(self, payload, user_id, now_str):
         birthdate = payload["officeholder"]["birthdate"].replace("-", "")
 
         # Extract and Process Birth Details
-        locality = str(
+        birthlocality = str(
```
ejjick: `birth_locality`?

web/se/edgeapi.py
```diff
@@ -65,7 +105,7 @@ def build_data(self, payload, user_id, now_str):
...
             payload["officeholder"]["birth_city"]).upper()
 
         birthcountry = str(
```
ejjick: `birth_country`

web/se/edgeapi.py
```diff
@@ -90,25 +130,7 @@ def build_data(self, payload, user_id, now_str):
...
-                postcode = address[3].lstrip()
-            else:
-                state = ""
-                postcode = ""
-                country = address[4].lstrip()
-        else:
-            country = ""
-            state = address[2].lstrip()
-            postcode = address[3].lstrip()
+        street, locality, state, postcode, country = validateaddress(address, update_id)  # noqa : E501
```
ejjick: `validate_address(...)`?

web/se/edgeapi.py
```diff
@@ -297,25 +319,7 @@ def build_data(self, payload, user_id, now_str):
             payload["shareholder"]["address"]
             ).upper().split(",")
```
ejjick: strange formatting here, nothing major if it passes

web/se/edgeapi.py
```diff
@@ -297,25 +319,7 @@ def build_data(self, payload, user_id, now_str):
...
-                postcode = address[3].lstrip()
-            else:
-                state = ""
-                postcode = ""
-                country = address[4].lstrip()
-        else:
-            country = ""
-            state = address[2].lstrip()
-            postcode = address[3].lstrip()
+        street, locality, state, postcode, country = validateaddress(address, update_id)  # noqa : E501
```
ejjick: could use multiple lines instead of `# noqa`

web/se/edgeapi.py
```diff
@@ -489,25 +493,7 @@ def build_data(self, payload, user_id, now_str):
             payload["shareholder"]["address"]
             ).upper().split(",")
```
ejjick: again, looks strange

web/se/tasks.py
```diff
@@ -6,8 +6,17 @@
...
+import json
+from django.conf import settings
+from datetime import datetime, timedelta
 
 logger = logging.getLogger(__name__)
+interval = datetime.utcnow() + timedelta(minutes=2)
+headers = {
+        'Content-Type': 'application/json',
+        'X-Auth-Edge':  settings.GETEDGE_API_KEY
+            }
```
ejjick: indentation

web/se/tasks.py
```diff
@@ -6,8 +6,17 @@
...
 
 
 @shared_task
```
ejjick: space

web/se/tasks.py
```diff
@@ -69,18 +86,87 @@ def lodgeupdateshareholder_address(payload, user_id, now, update_id):
...
+        checkresponselog(responseid, update_id)
+
+    else:
+        # Check Response Status after specified interval
+        checkresponsestatus.apply_async(
+            args=[responseid, update_id], eta=interval)
+
+
+@shared_task
+def checkresponselog(responseid, update_id):
```
ejjick: `check_response_log`?

web/se/tasks.py
```diff
@@ -69,18 +86,87 @@ def lodgeupdateshareholder_address(payload, user_id, now, update_id):
...
         responseid=responseid,
         status='LODGED')
+
+    # Check Response Status after specified interval
+    checkresponsestatus.apply_async(
+        args=[responseid, update_id], eta=interval)
+
+
+@shared_task
+def checkresponsestatus(responseid, update_id):
```
ejjick: `check_response_status(...)`?

web/se/edgeapi.py
troowala: you mean the starting with the ) ? 
its coz of the . that needs to go after ... that seems to be the best point to split to avoid noqa: E501

troowala:
Have made the requested changes. Please let me know once you are ok with this so it can be merged. Thanks

web/se/edgeapi.py
ejjick: Ah, for that you would add a `\` at the end of the line. I would put a `#noqa` when fixing the code is going to make it look strange. In this case I think the line would actually fit 80 characters width, otherwise the following looks better:

```python
address = str(payload["shareholder"]["address"]) \
          .upper().split(",")
```

or

```python
address = str(payload["shareholder"]["address"]) \
          .upper() \
          .split(",")
```

web/se/tasks.py
```diff
@@ -6,8 +6,17 @@
 from se.edgeapi import AddOfficeHolder, AddShareHolder, \
     UpdateShareHolderAddress, UpdateShareHolderShares
 import logging
+import requests
+import json
+from django.conf import settings
+from datetime import datetime, timedelta
 
 logger = logging.getLogger(__name__)
+interval = datetime.utcnow() + timedelta(minutes=2)
```
ejjick: This needs to be set in the function that calls it to update the time, otherwise the initial time will be used and the function will re-run immediately.

web/se/edgeapi.py
troowala: done. thx

web/se/tasks.py
troowala: done. thx

PR #167, opened 2019-01-08
Adding ability to select payment status. UX supports an input field f…

kimhansen:
…or partially paid shares, but value is not recorded.

### Pull Request check-list


web/se/static/js/app.js
```diff
@@ -62,7 +62,7 @@ var App = function() {
         query += " registeredAddress { street locality state postcode country } registeredAddressStartDate"
         query += " }"
         query += " } "
-        query += " shareholders ( organisationId: \"" + organisationId + "\" ) { id name address email sharetype { name } shares pricePerShare }"
+        query += " shareholders ( organisationId: \"" + organisationId + "\" ) { id name address email sharetype { name } shares pricePerShare paymentStatus }"
```
dpnova: No chance of fixing the way this string is created while you're here?

kimhansen: Yes, what is the best way of doing it?

PR #168, opened 2019-01-08
Feat/k new share issue select shareholder

kimhansen:
### Pull Request check-list


web/se/static/components/se/popup_edit_shareholder.tag
```diff
@@ -13,19 +10,33 @@ WTF
 
             <form>
                 <div class="modal-body">
-                        
-                    <div if={ !editing } class="input-field">
-                        <input id="name" ref="name" type="text" class="validate">
+
+                    <div if={ editing } class="input-field">
+                        <input id="name" ref="name" type="text" class="validate" readonly style="border: 0px; padding-left: 0px; margin-left: 0px;">
```
dpnova: inline css...........   :scream:

kimhansen: Yes, I think it is ok if it only stays for a week. The UX/UI will be improved next week.

PR #179, opened 2019-01-16
Intercom live chat integration

ejjick:


troowala:
all good ,  waitin on the issue kim needs to fix on the other PR rollback and then i suspect we can resume merging.

PR #186, opened 2019-01-21
Feat/k frontend changes for manual updates part2

kimhansen:
### Pull Request check-list


web/se/tests/test_schema.py
```diff
@@ -1733,6 +1733,49 @@ def test_addofficeholder_mutate_no_asic_update_success(
             AsicUpdate.objects.all().count()
         )
 
+    @patch("eventlog.tasks.log.apply_async", autospec=True)
+    def test_deleteofficeholder_mutate_no_asic_update_success(
```
troowala: test name should be deleteofficerholder_mutate_success 

no_asic_update is reserved for when the mutation is leveraging ,  if is_asic_update_enabled(e.organisation_id): which this mutation doesnt. (ref schema.py  lines 404 , to see an example)

web/se/tests/test_schema.py
```diff
@@ -1733,6 +1733,49 @@ def test_addofficeholder_mutate_no_asic_update_success(
...
+#        graphql_officeholder = executed['data']['deleteOfficeholder']['organisation']  # noqa: E501
+
+        self.assertEqual(
+            0,
+            Officeholder.objects.filter(
+                organisation_id="b280926e-8bae-4481-9d2c-d6a2d969400b"
+            ).count()
+        )
+
+        # Ensure no entries were written to ASIC Update Queue
```
troowala: this needs to check that an entry for a manual task is added instead of checking there are 0 items in the queue. also this is adding an entry , so this activity does add an entry to the queue which states task manual , status pending , so this should not be passing.

PR #206, opened 2019-01-30
Adding support for more shareholdertypes

kimhansen:
### Pull Request check-list


web/se/schema.py
```diff
@@ -60,6 +60,9 @@ class ShareRegistry(graphene.ObjectType):
 class ShareholderInputTypeEnum(graphene.Enum):
     BUSINESS = 'BUSINESS'
     PERSONAL = 'PERSONAL'
+    JOINT = 'JOINT'
```
troowala: Currently the edgeapi evaluates whether a shareholder is a business or not and if so then it accordingly includes the ACN. Would this be the same for Trust & SMSF ? 

Also joint shareholders are currently not implemented. What is the agreed method on implementing these @cballantyne @dpnova ? We need a chat about this.

PR #207, opened 2019-01-31
shareholder communication

ejjick:
### Pull Request check-list


ejjick:
Need to add validation and attachments functionality but I'd love to have this merged, been open too long

kimhansen:
Not receiving any emails though.

ejjick:
@kimhansen In local dev emails will be sent to `stdout`, but since it's handled by the Celery worker you won't see it. The fix for this is to shut down the worker and start your own.

```bash
docker-compose exec web bash

ps aux | grep celery
kill CELERYPID
celery -A tasks worker -l info
```

You can also check the Eventlog. This will only confirm that the task was run, not that the email was sent.

PR #218, opened 2019-02-12
recurring boto issues fix

troowala:
last few times there has been a new boto release it has broken integration testing and forces us to manually update the boto versions.

to address this we could change in requirements.txt

boto3==1.9.91
botocore==1.12.91

to 

boto3>=1.9.91
botocore>=1.12.91

while this may not be ideal as a new functionality by automatic upgrade might cause some other issue ,  however still breaks if we dont upgrade. so this might be a solution we should atleast try for a bit and see if we can avoid this issue.

### Pull Request check-list


kimhansen:
Would be good to have either Carl or D's feedback on this. I can't evaluate if it is a good idea or not.

dpnova:
Have discussed with @troowala - let's try it out and see how it goes.

PR #223, opened 2019-02-14
Feat/auth0 backend master

rdmrocha:
### Pull Request check-list


dpnova:
We decided against this approach to auth at the start of the project. Erik has some notes about it etc. So this is a reject from me.

kimhansen:
We are parking this for now.

PR #236, opened 2019-02-22
auth0 is now auto-refreshing

rdmrocha:
### Pull Request check-list


rdmrocha:
> Codewise seems fine. Testing isnt easy since we have to wait 2h until your session is expired. Will get to this when its merge!?

I tested it manually by changing the expire to 60 seconds on auth0 management console and commenting the re-auth part to make sure it will kick trigger a failure on our server, then with the re-auth part back and went for more than the 60s to validate our token gets refreshed. For regular users, you'd have to "wait 2h" and confirm that the platform never kicks you out

PR #237, opened 2019-02-22
fixing events, and HTML for best practices

rdmrocha:
When PR https://github.com/enhancedsociety/platform/pull/207 was merged, @kimhansen did not review the UI code for best practices and learnings previously shared in a small private session. 

As a result, this PR will keep the exact same functionality but without polluting the app with unnecessary events, by adding the riot tag in correct place, and normalising the code to use _let_ instead of _var_, accessors via jQuery and other minor tweaks.

Please bear in mind that both me and @pedroarelo (or at least one of us) should always review code that changes UI, being it JS, TAG, CSS, SCSS or HTML files.

<3


pedroarelo:
There's no UX for this feature and for sure should not be inside a popup. Wouldnt merge into master, would keep it in a branch until UX for this feature its there

PR #239, opened 2019-02-27
fixing date-pickers, kicking out users from details/edit/add if organ…

rdmrocha:
…isation changed

### Pull Request check-list


pedroarelo:
Don't agree with the current implementation of the Organisation changed event.
Every time you change organisation you should be routed back to the COMPANY details page, so the user is CERTAIN he is looking into a new company data, being thrown to the last page can became confusing. 
@kimhansen Whats your opinion on this? I guess my implementation is a better UX and has way less code.

kimhansen:
Yes, I agree. He should go to the Company details page. Once we have a dashboard, he might go there if it is clear enough that he is looking at the right company.

PR #245, opened 2019-03-01
Adding Google address autocomplete, with correct formatting.

kimhansen:
### Pull Request check-list


web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
+                address += val + format[counter++]
+            }
+        }
+    }
+    address += country;
+    return address;
+}
+
+function handleMessage(response, successMessage, errorMessage) {
+    if (response.errors != "") {
```
rdmrocha: if response.errors is an array of errors, we should be checking the length. otherwise line 140 might throw

web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
+var componentForm = {
+    street_number: 'short_name',
+    route: 'long_name',
+    locality: 'long_name',
+    administrative_area_level_1: 'short_name',
+    country: 'long_name',
+    postal_code: 'short_name'
+};
+
+var format = [" ", ", ", ", ", ", ", ", "]
```
rdmrocha: why var and not let? not critical, but just good practice

web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
+
+function getGoogleAddress(place) {
+    let address = ""
+    let country = ""
+    counter = 0
+    for (var i = 0; i < place.address_components.length; i++) {
```
rdmrocha: why var and not let? not critical, but just good practice

web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
+        var addressType = place.address_components[i].types[0];
```
rdmrocha: why var and not let? not critical, but just good practice

web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
+        if (componentForm[addressType]) {
+            var val = place.address_components[i][componentForm[addressType]];
```
rdmrocha: why var and not let? not critical, but just good practice

web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
-//         break
-//     case "lodgeupdateshareholder_address":
-//         text = "Update Shareholder Address"
-//         break
-//     default:
-//         text = "-"
-//     }
-//     return text
-// }
+var componentForm = {
```
rdmrocha: why var and not let? not critical, but just good practice

web/se/static/js/helpers/format.js
```diff
@@ -102,20 +102,44 @@ function formatASICAction(payload) {
...
+            } else {
+                address += val + format[counter++]
+            }
+        }
+    }
+    address += country;
+    return address;
+}
+
+function handleMessage(response, successMessage, errorMessage) {
```
rdmrocha: why are we even looking at errorMessage is none of the callers is sending a custom errorMessage. If not needed yet, please don't push as it might create confusion or legacy afterwards. If you plan on adding these custom error messages soon, please include the in the PR otherwise the method is  just doing an extra if every single time and the call stack needs to send more parameters every single time.

pedroarelo:
This conflicts with google chrome address fields.

screenshot 2019-03-01 at 11 44 38

web/se/static/js/helpers/format.js
kimhansen: yes, good point

web/se/static/js/helpers/format.js
kimhansen: error messages coming Monday/Tuesday

PR #247, opened 2019-03-02
Feat/k shareholders announcements improvements

kimhansen:
### Pull Request check-list


web/se/static/components/se/shareholders/dashboard/dashboard_activity.tag
```diff
@@ -1,21 +1,73 @@
...
 
     <script>
         let thizz = this;
+        thizz.state = '';
+        thizz.organisations = [];
+        thizz.messages = [];
+
+        // called when receiving data from server
+        thizz.onOrganisationDataReceived = function(data) {
+            if (thizz.organisations.length == 0) {
```
rdmrocha: data can have errors.you should check for that

web/se/static/components/se/shareholders/dashboard/dashboard_activity.tag
```diff
@@ -1,21 +1,73 @@
...
+                thizz.organisations = data.shareholderOrganisations
+            }
+            thizz.messages = data.shareholderMessage
```
rdmrocha: I would do something like thizz.messages = formatMessages(data.shareholderMessage) in order to keep formatMessages generic, reusable and not dependent on local context variables. There's nothing wrong with this approach in terms of functionality, but it's a best practice to only use global or context vars when needed.

kimhansen: Yes, it is better. I have fixed that.

PR #248, opened 2019-03-03
Creating announcements, instead of message popup.

kimhansen:
### Pull Request check-list


web/se/static/components/se/shareregistry/shareregistry_announcement_details.tag
```diff
@@ -0,0 +1,159 @@
...
+            <input show={ state != 'details' } type="submit" class="btn waves-effect" value="Send"></input>
+        </div>
+    </form>
+
+    <script>
+        let thizz = this
+        thizz.announcement = opts.announcement
+        thizz.state = opts.state
+        thizz.title = "Announcement Details"
+        thizz.canEditShareregistry = eval(riot.app.store.getCanEditShareregistry())
```
rdmrocha: why the eval? getCanEditShareregistry already returns a boolean so I don't really understand the need for this

web/se/static/components/se/shareregistry/shareregistry_announcement_details.tag
```diff
@@ -0,0 +1,159 @@
...
+            else {
+                thizz.refs.subject.value = ""
+                thizz.refs.body.value = ""
+            }
+            thizz.update()
+        }
+
+        thizz.submitCancel = function() {
+            if (thizz.state == 'add' || thizz.state == 'details') {
+                riot.app.events.triggerOrganisationChange(thizz.announcement.organisationId);
```
rdmrocha: why are we triggering an organisationChange on cancel?? are you trying to force an update on the other views?

web/se/static/components/se/shareregistry/shareregistry_announcements.tag
```diff
@@ -0,0 +1,106 @@
...
+
+    <div class="empty" if={ state == "" } show={ announcements.length <= 0 }>
+        List is empty
+    </div>
+
+    <script>
+        let thizz = this;
+        thizz.announcements = [];
+        thizz.announcement = null;
+        thizz.canEditShareregistry = eval(riot.app.store.getCanEditShareregistry())
```
rdmrocha: same as above: why eval?

web/se/static/components/se/shareregistry/shareregistry_announcements.tag
```diff
@@ -0,0 +1,106 @@
...
+
+        // called when receiving data from server
+        thizz.onOrganisationDataReceived = function(data) {
+            thizz.announcements = data.shareholderMessage
```
rdmrocha: data can contain errors and we should check for them

rdmrocha: I would do something like **_thizz. announcements = formatAnnouncements(data. shareholderMessage)_** in order to keep formatAnnouncements generic, reusable and not dependent on local context variables. There's nothing wrong with this approach in terms of functionality, but it's a best practice to only use global or context vars when needed.

web/se/static/components/se/shareregistry/shareregistry_announcement_details.tag
kimhansen: I will fix the eval in another PR, as it is not only for announcements.

web/se/static/components/se/shareregistry/shareregistry_announcement_details.tag
kimhansen: Yes, removed. Shouldn't be needed.

web/se/static/components/se/shareregistry/shareregistry_announcements.tag
kimhansen: Yes, I have improved this.

PR #250, opened 2019-03-04
Saving desired redirect path before app auth flow, route after auth s…

rdmrocha:
…uccessful. Also changed shareholders header to logout properly

### Pull Request check-list


rdmrocha:
it is working for me. If I'm logged in and try to access http://localhost:8000/se/shareholders/dashboard/activity for example, I get immediate access. If I'm logged out and I type that URI, I'll first be asked to login and then I'll land on the desired page. Same happens if I go to the root (shareregistry flow). Please double-check and post results

kimhansen:
Currently you can only enter the shareholders view if you are already logged in. If you are not logged in and put the link it will redirect to the shareregistry. If you are inside the shareholder view and logout and log in again, it will also open the shareregistry.

rdmrocha:
regarding logout, yes: once logged out, you'll be redirected to /se which will default you into /se/shareregistry upon login. This is because there is no flow for this, no distinction between entry points.

But if you try to go to /se/shareholders while logged out, you'll be taken straight to shareholders once authenticated. This latest commit fixes this issue. Logout will be addressed once a flow is created

PR #259, opened 2019-03-08
Feat/cb shareholding transaction

cballantyne:
### Pull Request check-list


web/se/schema.py
```diff
@@ -297,53 +311,77 @@ def mutate(root, info, transfer_shares_data=None):
...
-
-        # Increase shares of to shareholder
-        to_shareholder.shares = F('shares') + transfer_shares_data.quantity
-        to_shareholder.save()
-
-        # Log Event
-        log.apply_async(args=[
-            'se.shares.transfer', transfer_shares_data,
-            info.context.user.id, timezone.now()])
+        with transaction.atomic():
```
cballantyne: The only change here was adding 'with transaction.atomic():' and the block a bit further below - # Record shareholding transfer transaction. 
I had to change the indenting under transaction.atomic and it looks like that has confused the diff.

web/se/models.py
```diff
@@ -123,3 +123,20 @@ class ShareholderMessage(TimeStampedModel):
     user = models.ForeignKey(User, on_delete=models.CASCADE, db_index=True)
     subject = models.CharField(max_length=255)
     body = models.TextField()
+
+
+class ShareholdingTransaction(TimeStampedModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    shareholder = models.ForeignKey(
+        Shareholder, on_delete=models.CASCADE, db_index=True)
+    type = models.TextField()
```
rdmrocha: type field should use Django choices so the database model can limit and validate the values. Graphene will play along with these in terms of serialization/deserialization. ShareholdingTransactionTypeEnum will become unnecessary

cballantyne: Yeah I would like enums in the model but don't like how it translates down to the db level. Adding a new enum to type would result in migrations needing to be run. I have also run into issues in the past where enums in the tables (granted it was mysql) caused performance issues and data migration issues. I guess I have just been burned with enums in the db in the past and prefer them to be dumber, ie text fields. I don't think enums in the db is wrong just I shy away from it based on past experiences. @dpnova - do you lean a particular way?

cballantyne: On second thought. I will stop be a baby and just use the recommended approach by Django and put it in the model. It will help keep code cleaner.

cballantyne: Done. We are not returning the shareholding transactions yet so graphene using the types from the model won't be an issue yet. That will be the next thing after this is merged in.

PR #272, opened 2019-03-14
Investment history in shareholders & portfolio

pedroarelo:
Added investment history in Shareholders details;
Added Portfolio details;
Added investment history in Portfolio details;

Missing: getting the real investment history data
Missing: getting portfolio details not only by riot.opts
Missing: more fields in the portfolio details
Missing: delete functionality in portfolio details


web/se/schema.py
```diff
@@ -1586,6 +1586,13 @@ def is_org_partner(organisation_id, user_id):
     ).exists()
 
 
+def is_shareholder(organisation_id, user_id):
```
cballantyne: This needs a couple of unit tests for when true and when false.

kimhansen: Done, thanks!

PR #275, opened 2019-03-15
First experiment

kimhansen:
### Pull Request check-list


kimhansen:
It is still unclear how the valuation fits into the share registry. But what is clear is that it is not just another field like contact address.
Company valuation is a full feature in itself. It is a huge part of the ability to raise capital for a company. The data that is available from the history of transactions can provide great insights into the valuation of the company over time. Eg. each share issuance that is realised sets a valuation in time for the company.

PR #286, opened 2019-03-27
Close to working

kimhansen:
### Pull Request check-list


web/se/tests/test_schema.py
```diff
@@ -4685,3 +4685,67 @@ def test_is_shareholder_in_org(self):
...
+            'Not authorised',
+            executed['errors'][0]['message']
+        )
+        self.assertIsNone(
+            executed['data']['billingDetails']
+        )
+
+    # Should return Not authorised.
+    @patch("eventlog.tasks.log.apply_async", autospec=True)
+    def test_charge_payment_not_authorised(
```
cballantyne: Add another test for when charge payment is successful. It would check that the org table is updated with payment_subscription. You will need to mock stripe api call and I can help with that if you need.

web/se/tests/test_schema.py
```diff
@@ -4685,3 +4685,67 @@ def test_is_shareholder_in_org(self):
         shareholder.user.add(user)
 
         self.assertTrue(is_shareholder(org.id, user.id))
+
+    # Should return Not authorised.
+    def test_resolve_billing_details_not_authorised(
```
cballantyne: I would add a test when billing details are returned as well.

web/se/schema.py
```diff
@@ -1500,6 +1516,58 @@ def mutate(root, info, add_portfolio_company_data=None):
...
+            source=charge_payment_data.token_id,
+        )
+        logger.debug(f'customer : 'f'{customer}')
+
+        shareholders = ShareholderModel.objects.filter(
+            organisation__id=charge_payment_data.organisation_id
+        )
+        subscription = stripe.Subscription.create(
+            customer=customer.id,
+            items=[{'plan': 'plan_Elge7U0NaX9eyf', 'quantity': len(shareholders)}],  # noqa : E501
```
cballantyne: Don't hardcode plan key as this will definitely be different between environments. And may change over time as well.

kimhansen: Now getting the plans and checking if one with the right nickname exists (as the default).

web/se/tests/test_schema.py
kimhansen: Done

web/se/tests/test_schema.py
kimhansen: Done

PR #288, opened 2019-03-29
Extracting the address ASIC compliant address fields for Officeholders

rdmrocha:
Extracting the address ASIC compliant address fields, whenever possible. Only submitting when we have a match, otherwise we fallback to the string address like old times. 

This commit only applies to Officeholders.


rdmrocha:
Will create another branch based on this one for the shareholders. Will be named _**address_refactor_phase2.1**_

